PR #248, opened 2015-01-09
Generate styles/ directory instead of stylesheets/

nathansobo:
:warning: Don't merge until https://github.com/atom/atom/pull/4955 is released :warning:


nathansobo:
Thanks.

PR #260, opened 2015-01-16
Add --packages flag to apm list

### LINKED ISSUE #256: Add --no-themes or --packages flag to apm list
It would be really useful for `apm list` to be able to return a list of just *packages* and no themes. Currently it's possible to return a list of just *themes* and no packages, but not vice versa. I'm not sure which flag would make more sense, `--no-themes`, `--packages`, or something else. 

Something else might work better than those two options because it's not clear what the expected behavior would be if you were to call `apm list --themes --packages`, for instance. Does it throw an error? Does it return themes and packages (same as just running `apm list`)? Not sure.

mnquintana:
This implements my enhancement suggestion in #256 – it adds a `--packages` (or `-p`) flag for only listing packages. If combined with `--themes`, it'll just list everything, which seems fine to me.

Should I add a test for this? I noticed there aren't any tests for any of the flags, so I wasn't sure. (Maybe there should be?)

Either way, let me know! :smiley: 

(Resolves #256)

### DIFF
--- a/src/list.coffee
+++ b/src/list.coffee
@@ -27,6 +27,7 @@ class List extends Command
 
       Usage: apm list
              apm list --themes
+             apm list --packages
              apm list --installed
              apm list --installed --bare > my-packages.txt
              apm list --json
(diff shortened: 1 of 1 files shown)


src/list.coffee
```diff
@@ -69,6 +71,8 @@ class List extends Command
       manifest.name = child
       if options.argv.themes
         packages.push(manifest) if manifest.theme
+      else if options.argv.packages
+        packages.push(manifest) if not manifest.theme
```
kevinsawicki: Super minor, but usually we use `unless` instead of `if not`, mind updating?

mnquintana: No problem!

kevinsawicki:
Thanks for adding this :ship:

PR #262, opened 2015-01-19
Changing TextMate theme conversion to use current, non-deprecated selectors

brandonweiss:
The `TextMateTheme` class doesn't have tests, and when I tried to install the package from a folder (`npm install ~/Code/apm`) it tried to run `grunt prepublish` and complained the command couldn't be found, so I couldn't manually test it. But I think this should fix TextMate theme conversion.

delicate

### DIFF
--- a/src/text-mate-theme.coffee
+++ b/src/text-mate-theme.coffee
@@ -64,66 +64,75 @@ class TextMateTheme
 
   buildGlobalSettingsRulesets: (settings) ->
     @rulesets.push
-      selector: '.editor'
+      selector: 'atom-text-editor, :host'
       properties:
         'background-color': '@syntax-background-color'
         'color': '@syntax-text-color'
 
     @rulesets.push
-      selector: '.editor .gutter'
+      selector: 'atom-text-editor .gutter, :host .gutter'
       properties:
         'background-color': '@syntax-gutter-background-color'
         'color': '@syntax-gutter-text-color'
 
     @rulesets.push
-      selector: '.editor .gutter .line-number.cursor-line'
+      selector: 'atom-text-editor .gutter .line-number.cursor-line,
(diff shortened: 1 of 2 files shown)


nathansobo:
Thanks very much.

brandonweiss:
I think I may have broke something? Selecting text doesn't change the background color any more. I can't figure out why... the selector looks good, and the HTML elements/classes look like they should match the selector. But when I target the element in the inspector I can see that the style isn't applying.

brandonweiss:
Hmm, it might have something to do with the `.is-focused` class. This is what I see in the inspector.

Alt text

It looks like `.is-focused` is on the `atom-text-editor` element. But when I actually target an HTML element below it, `.is-focused` appears to be on a div inside of `atom-text-editor`.

My understanding of the shadow DOM is pretty minimal. Is that normal? Or is that a bug?

src/text-mate-theme.coffee
```diff
@@ -64,66 +64,75 @@ class TextMateTheme
...
-      selector: '.editor .search-results .marker.current-result .region'
+      selector: 'atom-text-editor .search-results .marker.current-result .region,
+                 :host .search-results .marker.current-result .region'
       properties:
         'border': '@syntax-result-marker-color-selected'
 
     @rulesets.push
-      selector: '.editor.is-focused .cursor'
+      selector: 'atom-text-editor.is-focused .cursor,
+                 :host.is-focused .cursor'
```
nathansobo: Oops, this needed to be `:host(.is-focused)`

nathansobo:
@brandonweiss Could you check out my comment about the `:host` selectors. I missed that in my review. Would be great if you could open another PR and ping me on it and I'll merge it right away.

PR #292, opened 2015-02-10
Validate semver ranges before publishing

### LINKED ISSUE #291: engines.atom semver not validated before publishing a version of a package
It's possible to publish a package version with an invalid `engines.atom` semver in the package.json, which causes problems like these https://github.com/atom/atom/pull/5417.

Wondering if we could+should validate that semver while publishing a package version (and perhaps other semvers listed in package.json?) and block publishing if the semver is invalid. :thought_balloon:

kevinsawicki:
Checks that the `engines.atom`, `dependencies`, and `devDependencies` semver ranges are valid in the `package.json` before allowing publishing to occur.

This fixes the case where a package can be published with an invalid engines field that won't be listed as compatible and so it actually can't be installed in any Atom version.

Closes #291

### DIFF
--- a/spec/publish-spec.coffee
+++ b/spec/publish-spec.coffee
@@ -1,0 +1,85 @@
+path = require 'path'
+fs = require 'fs-plus'
+temp = require 'temp'
+express = require 'express'
+http = require 'http'
+apm = require '../lib/apm-cli'
+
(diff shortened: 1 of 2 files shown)


izuzak:
:heart_eyes_cat:

PR #293, opened 2015-02-11
Fix missing symbol issue on Darwin, version bump, rewrite `apm` script.

geoff-codes:
:apple::facepunch::bug::arrow_up::art::fire::bathtub::package::arrow_right::v::octocat:

The long-standing issue of build failures due to missing symbols on Mac OS X [(#3184)](https://github.com/atom/atom/issues/3184) is the result of building modules with the 'system' node, as the build scripts require one. But the off-the-shelf 0.10.35 package that is download when apm is installed is not ABI compatible (This version of node is missing the `_node_module_register` symbol, which must be found in node binary when these modules are `dlopen()`d.

The simplest way to fix this ...is simply to bump node to v0.12.0. As it *has* now been released (and is a "stable" release), and because the apm test suite appears to have absolutely no issue with it, I have (hesitantly) done so in this PR. Hopefully this won't break things spectacularly.

ONE thing little thing this *does* break is the utilization of the `--harmony_collections` flag, whij has been removed as it is no longer needed. I have removed this flag in places where necessary.

Also, I rewrote your `apm` script. It kinda made me sad. This should be considerably more robust.

All the best,
-G

### DIFF
--- a/bin/apm
+++ b/bin/apm
@@ -1,40 +1,76 @@
-#!/bin/bash
+#!/bin/sh -e #  apm  - Atom Package Manager powered by https://atom.io
 
-set -e
+# node_flags="--harmony_collections"       ## Uncomment for node < 0.12.0.
 
-initialCwd=`pwd -P`
+unset -f cd 2>/dev/null ||: alias cd=\cd   ## Kill heinous cd hooks (rvm).
 
-apmPath=$0
-builtin cd "`dirname "$apmPath"`"
-binDir=`basename "$apmPath"`
+[ -"$BASH_VERSION"- != -""- ] && set -o posix  ## Try to get a sane shell.
+[ -"$ZSH_VERSION"-  != -""- ] && emulate sh && setopt shwordsplit
 
-# Detect CPU arch.
-archs=`uname -m`
-case "$archs" in
-  i?86) archs='ia32' ;;
-  x86_64) archs='x64' ;;
-  *) archs='arm' ;;
-esac
+#===------------------------------------------------------- readlink() -===##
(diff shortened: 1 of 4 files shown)


geoff-codes:
Also: https://github.com/atom/atom-shell/issues/1070

geoff-codes:
And when this is taken care of, the next issue we run into building on Darwin is:
[**zcbenz/base-minimal**: Add missing `<pthread.h>` (Pull #4)](https://github.com/zcbenz/base-minimal/pull/4)

bin/apm
```diff
@@ -1,40 +1,76 @@
...
-do
-  binDir=`readlink "$binDir"`
-  builtin cd "`dirname "$binDir"`"
-  binDir=`basename "$binDir"`
-done
-
-binDir=`pwd -P`
-builtin cd "$initialCwd"
-"$binDir/$nodeBin" --harmony_collections "$binDir/../lib/cli.js" "$@"
+#!/bin/sh -e #  apm  - Atom Package Manager powered by https://atom.io
```
kevinsawicki: What issues do these changes fix?

geoff-codes:
Eh its mostly just shell and platform portability. Probably not worth it.

kevinsawicki:
> Eh its mostly just shell and platform portability. Probably not worth it.

I'm unclear why you closed this, it seems like the node 0.12 change is still very important to have in there.

Were the apm script changes related at all to supporting node 0.12?

geoff-codes:
> I'm unclear why you closed this, it seems like the node 0.12 change is still very important to have in there.

Sorry, I hadn't realized how close you were on that, so I closed it especially since

> Were the apm script changes related at all to supporting node 0.12?

... they were not.

PR #314, opened 2015-02-27
Make `keytar` an optional dependency

naholyr:
As keytar's build can often fail (see atom/node-keytar#7 and atom/node-keytar#18) and it's already made optional on top of `src/auth.js`, why not make the dependency optional?
With current config it will fail at `npm install` whereas apm could completely work without it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
     "colors": "~0.6.1",
     "wordwrap": "0.0.2",
     "fs-plus": "2.x",
-    "keytar": ">=2.1 <3",
     "read": "~1.0.5",
     "q": "~0.9.7",
     "plist": "git+https://github.com/nathansobo/node-plist.git",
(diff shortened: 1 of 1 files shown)


kevinsawicki:
> why not make the dependency optional?

One downside to this (I'm assuming) would be our Atom CI builds on all platforms would no longer fail on any keytar build issues they encounter.

So we could end up redistributing a release of Atom that does not contain keytar which would be really unfortunate on Windows and Mac where it works well out of the box.

Is there a way to make optional dependencies on a per platform basis?

naholyr:
But in which situation would it fail to build but still worth including? :D
Not sure how you can achieve platform-specific dependencies. Will dig.

Le ven. 27 févr. 2015 18:14, Kevin Sawicki <notifications@github.com> a
écrit :

bolinfest:
Because Atom has no API for packages to secure information, it's critical for Nuclide that keytar is available in Atom, even if it is only there indirectly due to apm.

damieng:
Decided not to make this optional for the reasons stated above.

PR #316, opened 2015-02-28
Bump year in template licenses to 2015

benjaminparnell:
I noticed this while generating some packages with `apm init`. Maybe its worth making the year dynamic so its always the current year? Happy to make this change.


bouzuya:
:+1:

kevinsawicki:
Thanks for updating these

PR #327, opened 2015-03-25
Update README.md

tarpdalton:
update .apmrc file path.


kevinsawicki:
The current path is correct, config settings should go in `~/.atom/.apmrc`

PR #330, opened 2015-03-30
:racehorse: Exclude installed packages from stars --install

### LINKED ISSUE #170: apm stars --install is very slow
I am finding the apm star(s) commands great for helping keep my package usage in sync between machines, but I find it very slow when running apm stars --install.

It seems to re-install all my starred packages every time.

It would be great if the behavior was able to be controlled so that the following was possible as a single and/or individual set of switches to the command line:
- Install missing starred packages
- Install outdated starred packages
- Ignore installing/updating starred packages that are currently disabled

Thanks

James

AsaAyers:
Fixes #170 

I didn't add a spec as I can't run the tests on my machine. It doesn't give me any useful information.

selection_002

edit: I just verified it works by running `./bin/apm stars --install`

### DIFF
--- a/src/stars.coffee
+++ b/src/stars.coffee
@@ -70,15 +70,13 @@ class Stars extends Command
         dev: true
         bare: true
 
-    list.listDevPackages listOptions, (error, devPackages) ->
-      devPackages = devPackages.map(toName)
-      list.listUserPackages listOptions, (error, userPackages) ->
-        userPackages = userPackages.map(toName)
+    list.listUserPackages listOptions, (error, userPackages) ->
+      userPackages = userPackages.map(toName)
 
-        commandArgs = packages.map(toName).filter (name) ->
-          ( name not in devPackages ) and ( name not in userPackages )
+      commandArgs = packages.map(toName).filter (name) ->
+        return name not in userPackages
 
-        new Install().run({commandArgs, callback})
+      new Install().run({commandArgs, callback})
 
   logPackagesAsJson: (packages, callback) ->
     console.log(JSON.stringify(packages))
--- a/src/stars.coffee
+++ b/src/stars.coffee
@@ -5,6 +5,7 @@ Command = require './command'
 config = require './apm'
 Install = require './install'
 Login = require './login'
+List = require './list'
 request = require './request'
 tree = require './tree'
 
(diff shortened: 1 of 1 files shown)


Fank:
This is awesome :+1:
FYI, this also fixes an issue with "connection establishing problems":
Im working behind a proxy, sometimes the proxy connection will not establish, so the package will not be installed and apm will exit.
The problem is if you got a long stared list, you will be unable to install the list and need to switch to step by step installation.
And now you could easily install all stared packages by running `apm stars --install` multiple times until all packages are installed.

src/stars.coffee
```diff
@@ -60,8 +61,24 @@ class Stars extends Command
...
+    toName = ({name}) -> name
+
+
+    list = new List()
+    listOptions =
+      argv:
+        dev: true
+        bare: true
+
+    list.listDevPackages listOptions, (error, devPackages) ->
```
izuzak: Can you share some thoughts on why both user and dev packages should be checked? I _think_ I would have expected only user packages to be checked, because that's how these packages will be installed.

izuzak:
Thanks for working on this, @AsaAyers! :tophat: I also think this would be useful.

> I didn't add a spec as I can't run the tests on my machine. It doesn't give me any useful information.

That's strange, I just tried running test on your branch and they passed. Are you still seeing those problems? Does running `npm test --verbose` give you any more information?

It would be great if you could add some tests for this -- it would make us feel more confident about merging and prevent the feature from breaking in the future. 

Also cc @kevinsawicki and @thedaniel who'll probably have more :thought_balloon:.

src/stars.coffee
AsaAyers: I misunderstood how dev packages worked. I thought it was a problem to have the same package in both. I'll update this to not check dev packages.

AsaAyers:
The specs are working now. I created this PR while I was trying to use Windows. Now that I'm back to Linux everything actually works.

src/stars.coffee
```diff
@@ -60,8 +61,22 @@ class Stars extends Command
...
+        dev: true
+        bare: true
+
+    list.listUserPackages listOptions, (error, userPackages) ->
+      userPackages = userPackages.map(toName)
+
+      commandArgs = packages.map(toName).filter (name) ->
+        return name not in userPackages
+
+      new Install().run({commandArgs, callback})
```
AsaAyers: How can I get a reference to `Install` or `Stars` from the spec? I'm thinking repeat the steps for the ` "when the install flag is specified" -> it "installs all of the stars"`, then run the same command a second time and verify that the list is empty by the time it's passed over to `new Install().run(...)`

izuzak: @AsaAyers That's a good question. I _think_ you could use [this approach](https://github.com/AsaAyers/apm/blob/e144f99e01be02322d07f0799bb34b6384c7882f/spec/develop-spec.coffee#L25) to spy on `Install.run()` (but still allow the call to go through). What do you think?

firstdoit:
Hey, everyone! What's missing from this PR exactly? I can give a hand.

AsaAyers:
I'm making https://github.com/pulls more useful to me today by closing my old PRs.

PR #345, opened 2015-04-23
Update node version to 0.22.3

suda:


src/apm.coffee
```diff
@@ -61,7 +61,7 @@ module.exports =
     process.env.ATOM_API_URL ? 'https://atom.io/api'
 
   getNodeVersion: ->
-    process.env.ATOM_NODE_VERSION ? '0.22.0'
+    process.env.ATOM_NODE_VERSION ? '0.22.3'
```
kevinsawicki: @zcbenz does this need to be updated?

zcbenz: No, we only need to update node version when the minor version of Electron changes.

suda: It looks like it has to be updated as it's used to build/search native modules: 
screenshot 1
(`serialport` module was installed using `Update Package Dependencies: Update` command)

PR #356, opened 2015-05-06
Don't limit by operating system type. Default to reasonable node binary

pusateri:
archs is unused in this file. Remove.
Don't limit use by operating system. Set correctly for Windows and pick reasonable default for all else.
This fixes a problem on FreeBSD.

### DIFF
--- a/bin/apm
+++ b/bin/apm
@@ -8,24 +8,11 @@ apmPath=$0
 builtin cd "`dirname "$apmPath"`"
 binDir=`basename "$apmPath"`
 
-# Detect CPU arch.
-archs=`uname -m`
-case "$archs" in
-  i?86) archs='ia32' ;;
-  x86_64) archs='x64' ;;
-  *) archs='arm' ;;
-esac
-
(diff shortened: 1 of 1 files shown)


kevinsawicki:
Awesome, thanks for this.

pusateri:
Sorry for the expr problem. I see you fixed it by adding back a Darwin test. That will work for now but long term, I would like to rewrite this to not test for OS type.

This usage of expr is not posix standard. But I think bash syntax will work and be portable. I’ll submit another patch after I test it.

Thanks,
Tom

kevinsawicki:
> That will work for now but long term, I would like to rewrite this to not test for OS type.

Sounds good, thanks.

pusateri:
Shouldn’t the Jenkins CI have failed?

Thanks,
Tom

kevinsawicki:
> Shouldn’t the Jenkins CI have failed?

Good point.

The specs run by calling the commands internally, not directly spawning them as sub-processes so the `bin/apm` file isn't exercised in the specs currently.

At least one test should be added though to exercise it from the outside to verify it does spawn correctly.

PR #359, opened 2015-05-06
Use bash substrings instead of expr non-posix substrings for portability.

pusateri:
I've tested this on OS X, debian, and FreeBSD. I don't have the ability to test on Windows.


kevinsawicki:
Awesome, thanks!

Issue #282, opened 2015-02-02
Build Fails on FreeBSD Error: incorrect header check   at Zlib


mike-pt:
> node ./script/download-node.js

events.js:72
        throw er; // Unhandled 'error' event
              ^
Error: incorrect header check
  at Zlib._binding.onerror (zlib.js:296:17)

mike-pt:
Any idea how to fix/work around it?

pusateri:
This throw er; turns out to be the fact that apm is trying to download a binary package of node to run and freebsd isn't in the list of built packages.

http://nodejs.org/dist/v0.10.35/

% node ./script/download-node.js
http://nodejs.org/dist/v0.10.35/node-v0.10.35-freebsd-x64.tar.gz

If I'm running the script with node, why is the script downloading node?

kevinsawicki:
Closed by #360

mike-pt:
This doesn't fix the issue completely for freebsd, it simply help understand better what happens (404 downloading node).

In practice the zlib issue is gone but compilation still fails for freebsd, and node is already installed anyway, if node is installed why even try to download it again?

If the error is 404 but node is installed apm build should proceed.

mike-pt:
Or the other way around... (test if node is installed first and only if not attempt to download it)

mike-pt:
Sorry for the spam, but I was looking at the what the ports are getting as a distfile, and it seems its "node-v0.xx.x.tar.gz", so maybe a good way to handle it would be to just check if its freebsd and download the tar.gz 

So basically if its "freebsd" remove "process.platform", I could make a PR for that change if it sounds reasonable?

pusateri:
Yes, this is a baby step. I hope we can make it fall back to the native version. In the mean time, use this patch:
https://gist.github.com/pusateri/a5283313fdbef9922ce6#file-package-json-diff

mike-pt:
Sounds ok, I'm my case I had to use ln -s instead of creating a hard link but this is a nice workaround for now.

pusateri:
Yeah, mine is one big ZFS tank. I'll change that diff to use a symbolic link.
After building amp, checkout atom and apply this patch:
https://gist.github.com/pusateri/eb98a25e3ae45309aed7#file-freebsd-atom-diff

This will point atom at the patched apm.
minidump is for creating crash dumps. We can live without it for now.
And 'rm build/tasks/dump-symbols-task.coffee' for now.

mike-pt:
[...]

PR #360, opened 2015-05-07
Catch 404 not found case and quit stream before passing zlib empty file

pusateri:
Currently, if you try to download node for unsupported architecture/operating system, you get an exception because it passes a null stream to zlib to decompress.

events.js:72
        throw er; // Unhandled 'error' event
              ^
Error: incorrect header check
    at Zlib._binding.onerror (zlib.js:296:17)

This change results in and error being printed:
download not found: http://nodejs.org/dist/v0.10.35/node-v0.10.35-freebsd-x64.tar.gz

### DIFF
--- a/script/download-node.js
+++ b/script/download-node.js
@@ -37,8 +37,14 @@ var downloadTarballAndExtract = function(url, location, callback) {
   });
   stream.on('end', callback.bind(this, tempPath));
   stream.on('error', callback);
-  request.createReadStream({url: url}, function(requestStream) {
-    requestStream.pipe(zlib.createGunzip()).pipe(stream);
+  request.createReadStream({url: url}, function(res) {
+    res.on('response', function (resp) {
+      if (resp.statusCode == 404) {
+        console.error('download not found:', url);
+        process.exit(1);
+      }
+      res.pipe(zlib.createGunzip()).pipe(stream);
+    });
   });
 };
 
--- a/script/download-node.js
+++ b/script/download-node.js
@@ -37,13 +37,13 @@ var downloadTarballAndExtract = function(url, location, callback) {
   });
   stream.on('end', callback.bind(this, tempPath));
   stream.on('error', callback);
-  request.createReadStream({url: url}, function(res) {
-    res.on('response', function (resp) {
-      if (resp.statusCode == 404) {
+  request.createReadStream({url: url}, function(requestStream) {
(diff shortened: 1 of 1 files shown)


kevinsawicki:
I believe this would fix #282

pusateri:
Yes, and it should make the problem obvious for the Linux arm guys too. I’m not sure if the coffeescript version of request is used vs. the library version of request if this will act differently and I’m not sure how to test that.

What are your thoughts on something like:

	If no download is available and the node used to run script/download-node.js is the same major version as specified in the call to downloadNode(), then fall back to copying it?
		cp -p `which node` bin/node

In fact, it seems like the cp could be done first if the version matches and it if doesn’t, then do the download. I’m still figuring out how all of this works but the fact that node is running the script to download node seems like there could be a shortcut here under some circumstances.

Thanks!
Tom

script/download-node.js
```diff
@@ -37,8 +37,14 @@ var downloadTarballAndExtract = function(url, location, callback) {
   });
   stream.on('end', callback.bind(this, tempPath));
   stream.on('error', callback);
-  request.createReadStream({url: url}, function(requestStream) {
-    requestStream.pipe(zlib.createGunzip()).pipe(stream);
+  request.createReadStream({url: url}, function(res) {
```
kevinsawicki: Mind keeping the variable as `requestStream`, having both `res` and `resp` variables is a little confusing.

pusateri: Agreed. I will redo it. Is there a better way to get feedback on a patch without doing a pull request?

Thanks,
Tom

kevinsawicki:
> I’m not sure if the coffeescript version of request is used vs. the library version of request if this will act differently and I’m not sure how to test that.

There shouldn't be any differences and I don't think you need to test that.

> If no download is available and the node used to run script/download-node.js is the same major version as specified in the call to downloadNode(), then fall back to copying it?

Yeah, that might be something work exploring, but it would be nice to just get downloads working on all platforms where Atom works.

> I’m still figuring out how all of this works but the fact that node is running the script to download node seems like there could be a shortcut here under some circumstances.

apm ships with Atom where node can't be assumed to be installed so this is really just a tool to bundle node with Atom and apm so that apm works from the terminal.

It isn't intended that someone ever runs `npm install atom-package-manager` directly.

kevinsawicki:
Thanks for fixing this.

PR #361, opened 2015-05-08
Use /usr/bin/env bash in hashbang

mnquintana:
From https://github.com/atom/apm/issues/282#issuecomment-100289741


kevinsawicki:
:+1:

PR #369, opened 2015-06-01
Don't install or show search results for deprecated packages

kevinsawicki:
Use a list of known packages with deprecations to prevent installation of packages not compatible with Atom 1.0

Only affects the `apm search` and `apm install` commands.

Refs https://github.com/atom/atom/issues/6867

/cc @benogle


kevinsawicki:
Travis is having issues: https://www.traviscistatus.com/incidents/2kz853x98kr9

Tests pass for me locally :green_apple:

Issue #372, opened 2015-06-08
Run `prepublish` script from package.json on apm publish


Glavin001:
Applicable `apm` code: https://github.com/atom/apm/blob/6c94d96f5d640a51d49a41e2e1d1aa47310516e8/src/publish.coffee

I would like to use the `prepublish` script in `package.json` file, as described by `npm`, to generate my documentation before publishing. See https://github.com/Glavin001/atom-beautify/blob/master/package.json#L150

This currently does not work / is not a feature. Is this something that is of interest to others? Would a pull request be accepted? What other considerations does the community have, before / if I were to develop a Pull Request?

Thanks!

Glavin

kevinsawicki:
> Is this something that is of interest to others?

Yeah, this issue has come up before as something that would be great to eventually have first-class support for.

> Would a pull request be accepted?

:+1: 

> What other considerations does the community have, before / if I were to develop a Pull Request?

There is a `publish-spec.coffee` already so I'd say add new specs there, and add a simple package fixture that includes a `prepublish` script.

Also this will require doing a `git commit` after running the script right? Since the built assets will need to be part of the repository before the tag is pushed and the `atom.io` API is hit.

Glavin001:
Here is `publish` from `npm`: https://github.com/npm/npm/blob/180da67c9fa53103d625e2f031626c2453c7ebcd/lib/publish.js#L74
And `publish` specs: https://github.com/npm/npm/blob/180da67c9fa53103d625e2f031626c2453c7ebcd/test/tap/prepublish.js

> Also this will require doing a git commit after running the script right? Since the built assets will need to be part of the repository before the tag is pushed and the atom.io API is hit.

I think you're right. 

`npm` does not deal with `git` and just flows down the lifecycle: https://github.com/npm/npm/blob/180da67c9fa53103d625e2f031626c2453c7ebcd/lib/publish.js#L74-L77

So would `apm` run `prepublish` and then check if changes to staging area and commit if available?
What would the commit message be? Would there be a way to let the contents be included under the `Prepare # release` commit? See https://github.com/atom/apm/blob/b47423db2dd093404cab80d752793cdf91986111/src/publish.coffee#L56

`npm version` requires that it be a clean Git working directory.

From https://docs.npmjs.com/cli/version:

> If run in a git repo, it will also create a version commit and tag, and fail if the repo is not clean.

Glavin001:
[...]

PR #376, opened 2015-06-15
WIP: Add prepublish and postpublish scripts support to apm publish

Glavin001:
See https://github.com/atom/apm/issues/372

### DIFF
--- a/src/publish.coffee
+++ b/src/publish.coffee
@@ -286,6 +286,95 @@ class Publish extends Command
       # Just fall through if the name is empty
       callback()
 
+  # Check if Git working directory is clean
+  # callback - The callback function to invoke when done with an error as the
+  #            first argument, clean status as second argument.
+  #            The clean status is true if the Git working directory is clean.
+  checkCleanGit: (callback) ->
+    config.getSetting 'git', (gitCommand) =>
+      gitCommand ?= 'git'
+      # Add all changes to Git staging area
+      @spawn gitCommand, ['status', '--porcelain'], (code, stderr='', stdout='') =>
+        unless code is 0
+          @logFailure()
+          addOutput = "#{stdout}\n#{stderr}".trim()
+          return callback("`git status --porcelain` failed: #{addOutput}")
+
+        lines = stdout.trim()
+          .split("\n")
+          .filter((line) ->
+            line.trim() && !line.match(/^\?\? /)
+          ).map((line) -> line.trim() )
+        if lines.length
+          # Not clean
+          callback(null, false)
+        else
+          # is clean
+          callback(null, true)
+
+  # Run prepublish script for package
+  #
+  prepublishPackage: (pack, version, callback) ->
+    # Check if prepublish script exists
+    if pack.scripts?.prepublish?
+      message = "Prepublish #{pack.name} for #{version} release"
+      process.stdout.write(message)
+      # Run prepublish script with npm
+      runArgs = ['run', 'prepublish']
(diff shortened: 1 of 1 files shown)


Glavin001:
Would some of the maintainers of Atom / apm be able to give me some points on the questions above? Thanks in advance!

kevinsawicki:
> Should I run git init to make the temporary directory a Git repository and then it would no longer error with Package must be in a Git repository before publishing...?

> How do I get past Package must pushed up to GitHub before publishing?

For both of these, you can create a repository locally that satisfies these conditions, and check it in to `spec/fixtures` and then you can just copy it to a temp folder and use it in the specs.

> To confirm, once I get past the above, I can safely run publish and it will not attempt to publish to the global apm API because ATOM_API_URL is set to http://localhost:3000/api?

Yup, as long as `ATOM_API_URL` is set to a localhost address no calls will go out to `atom.io`

Glavin001:
> > How do I get past Package must pushed up to GitHub before publishing?

> For both of these, you can create a repository locally that satisfies these conditions, and check it in to spec/fixtures and then you can just copy it to a temp folder and use it in the specs.

So in the case of the Package being on GitHub, wouldn't a new tag be pushed every single time the tests are successfully run?
See https://github.com/Glavin001/apm/blob/master/src/publish.coffee#L445-L448 calling `@pushVersion` and `@waitForTagToBeAvailable`.
This does not sound ideal for testing to me.
Also, I believe that it would even error on the second run because the Git repo in `spec/fixtures` has the tag, say v1.0, and that version tag has already been previously pushed to GitHub, so it will fail this second time.

I was thinking along the lines of giving the repo a fake `upstreamUrl` and then mocking [`pushVersion`](https://github.com/Glavin001/apm/blob/master/src/publish.coffee#L70) and [`waitForTagToBeAvailable`](https://github.com/Glavin001/apm/blob/master/src/publish.coffee#L86) functions such that they callback successfully without actually sending requests to GitHub.
Thoughts on this implementation?

bronson:
Question about potential usage...  Any chance I could write a prepublish script that would use git log to fake up a new entry in the CHANGELOG and open it in an editor?  Then, when I save the changelog engry, the package would be published as usual.

Glavin001:
@bronson: Absolutely! This will wait until the prepublish script returns with it's exit code and will only continue afterwards if it was successful. That's a good use case. Personally, I will be using this for running a task for documentation generation before publishing.

iam4x:
Hey @Glavin001 any update on this? Do you need help?

steelbrain:
Just a side question: Why are we using the callback-hell when promises are available?

Glavin001:
@iam4x : I have not made any progress since my last commit. Any help would be appreciated. Unfortunately, I am now studying for my university final exams and will not have much time to work on this in the next few weeks.

@steelbrain : I would definitely prefer to use promises! :+1:

danielbayley:
Shouldn't this be expanded to include _all_ relevant hooks? For example `preversion`/`postversion`…

idleberg:
I [have noticed](https://code.visualstudio.com/docs/tools/vscecli#_prepublish-step) that Visual Studio Code prefixes its `prepublish` script in the manifest. I wondered if it should be considered a good practice to use `atom:prepublish` instead to differentiate from npm's script – not that I could think of a real life conflicting situation.

aminya:
Could you fix the conflicts? This seems like a very critical feature that is missing from apm.

Issue #391, opened 2015-07-03
apm unlink <package_path> doesn't work for numeric package names


izuzak:
I have a linked package, called `123453543534` :stuck_out_tongue_winking_eye::

```
$ apm linked
/Users/izuzak/.atom/dev/packages (0)
└── (no links)
/Users/izuzak/.atom/packages (1)
└── 123453543534 -> /Users/izuzak/github/123453543534
```

`apm unlink` doesn't work if I pass the package's name (it works for other packages which have at least one non-digit character):

```
$ apm unlink 123453543534

path.js:313
        throw new TypeError('Arguments to path.resolve must be strings');
              ^
TypeError: Arguments to path.resolve must be strings
    at Object.exports.resolve (path.js:313:15)
    at Unlink.module.exports.Unlink.unlinkPackage (/Applications/Atom.app/Contents/Resources/app/apm/lib/unlink.js:91:23)
    at Unlink.module.exports.Unlink.run (/Applications/Atom.app/Contents/Resources/app/apm/lib/unlink.js:130:21)
    at Object.module.exports.run (/Applications/Atom.app/Contents/Resources/app/apm/lib/apm-cli.js:226:32)
    at Object.<anonymous> (/Applications/Atom.app/Contents/Resources/app/apm/lib/cli.js:6:7)
    at Object.<anonymous> (/Applications/Atom.app/Contents/Resources/app/apm/lib/cli.js:17:4)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
```

Unlinking with `apm unlink .` works fine :white_check_mark: :

```
$ apm unlink .
Unlinking /Users/izuzak/.atom/packages/123453543534 ✓
```

PR #392, opened 2015-07-03
Stringify argument to apm link / unlink

mnquintana:
Fixes https://github.com/atom/apm/issues/391

/cc @atom/feedback

### DIFF
--- a/spec/link-spec.coffee
+++ b/spec/link-spec.coffee
@@ -146,9 +146,8 @@ describe 'apm link/unlink', ->
       runs ->
         expect(fs.existsSync(path.join(atomHome, 'packages', path.basename(numericPackageName)))).toBeTruthy()
         expect(fs.realpathSync(path.join(atomHome, 'packages', path.basename(numericPackageName)))).toBe fs.realpathSync(numericPackageName)
-        callback.reset()
 
-      runs ->
+        callback.reset()
         apm.run(['unlink', numericPackageName], callback)
 
       waitsFor 'unlink to complete', ->
--- a/spec/link-spec.coffee
+++ b/spec/link-spec.coffee
@@ -131,3 +131,30 @@ describe 'apm link/unlink', ->
         expect(fs.existsSync(path.join(atomHome, 'dev', 'packages', path.basename(packageToLink1)))).toBeFalsy()
         expect(fs.existsSync(path.join(atomHome, 'packages', path.basename(packageToLink2)))).toBeFalsy()
         expect(fs.existsSync(path.join(atomHome, 'packages', path.basename(packageToLink3)))).toBeFalsy()
+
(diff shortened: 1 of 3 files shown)


mnquintana:
Anybody wanna review this? :bow: @atom/feedback

spec/link-spec.coffee
```diff
@@ -131,3 +131,30 @@ describe 'apm link/unlink', ->
...
+      runs ->
+        apm.run(['link', numericPackageName], callback)
+
+      waitsFor 'link to complete', ->
+        callback.callCount is 1
+
+      runs ->
+        expect(fs.existsSync(path.join(atomHome, 'packages', path.basename(numericPackageName)))).toBeTruthy()
+        expect(fs.realpathSync(path.join(atomHome, 'packages', path.basename(numericPackageName)))).toBe fs.realpathSync(numericPackageName)
+        callback.reset()
```
thomasjo: Extremely minor, but I'd move to this reset call into the `runs` block below (the unlink one).

thomasjo:
Besides my one :art: comment, this looks good to me!

src/link.coffee
```diff
@@ -29,6 +29,8 @@ class Link extends Command
     {callback} = options
     options = @parseOptions(options.commandArgs)
 
+    options.argv._[0] = if options.argv._[0] then String(options.argv._[0])
```
kevinsawicki: Instead of assigning back to this array, what do you think about pulling out a variable that incorporates defaulting and stringifying into its own line.

Something like:

```coffee
packagePath = options.argv._[0]?.toString() ? '.'
linkPath = path.resolve(process.cwd(), packagePath)
```

mnquintana: Oh yeah, that looks way better :+1:

mnquintana: Done!

spec/link-spec.coffee
```diff
@@ -25,9 +25,8 @@ describe 'apm link/unlink', ->
       runs ->
         expect(fs.existsSync(path.join(atomHome, 'packages', path.basename(packageToLink)))).toBeTruthy()
         expect(fs.realpathSync(path.join(atomHome, 'packages', path.basename(packageToLink)))).toBe fs.realpathSync(packageToLink)
-        callback.reset()
 
-      runs ->
+        callback.reset()
```
kevinsawicki: I combined the consecutive `runs` blocks in this spec file for clarity. Usually we don't have multiple `runs` block after each other since it essentially the same as having it all in a single `runs` block.

kevinsawicki:
Travis having issues, https://www.traviscistatus.com/incidents/khzk8bg4p9sy

I ran the specs locally and they passed, shipping this, thanks @mnquintana

Issue #358, opened 2015-05-06
Add `apm disable / enable` command


mnquintana:
[There is currently no way to disable packages from the command-line](https://github.com/atom/language-html/issues/59#issuecomment-99542896) - we should add an `apm disable` command that can take one or more package names and disable them in Atom (as well as its counterpart, `apm enable`).

brennanMKE:
:+1:

bronson:
It's possible to do manually: https://github.com/atom/atom/issues/5987#issuecomment-84129975

But, I agree, :+1: for `apm disable`.

bronson:
@mnquintana now there's a way.  Chance of a review for #400?

dfang:
+1

echohack:
+1 please

nmccready:
for the love +1 also doing this in bulk would be great

bronson:
@nmccready the PR (#400) accepts multiple packages on the command line:
```
apm disable metrics exception-reporting
```

is this what you mean by bulk?

nmccready:
yep ty

nmccready:
Any suggestions how I could use `xargs` from `apm list` to disable all installed packages (not default) . I am unable to get the editor to open with latest update.

Sirlon:
+1

pmsm1100:
+1

bronson:
@nmccready sorry, I don't know of a good way.  In theory it ought to look something like this:

```sh
apm disable $(apm list --installed --bare)
```

But that doesn't work because --bare still includes "@_version_" after each package name.

I like the idea of a --really-bare option that would make that work.  Probably worth opening a new issue?

(Or, maybe apm should just remove the "@_version_" from --bare's output.  That's what I picture by "bare".  :)

nmccready:
@bronson , no worries; I had to blow everything away and start from scratch. Atom keeps breaking thing with all the updates lately.

danielbayley:
@bronson `apm list -ib | sed s/@.*//g` will give you a plain text list of all packages.

`apm list -ibd false | sed s/@.*//g` will do the same, excluding disabled packages.

bronson:
@nmccready Yes, it's been a particularly rough patch over the past month or two.  Hoping things will settle down now that Electron and Node are updated.

@danielbayley agreed, that does the job here.  But I think it's a bad sign when a command line program can't ingest its own output.  It would be nice to have a proper fix.

[...]

PR #400, opened 2015-07-19
Add enable and disable commands

bronson:
Addresses https://github.com/atom/apm/issues/358 and https://github.com/atom/atom/issues/5987.

There's some code duplication but that seems to be apm's style: commands and tests are mostly self-sufficient.  I wanted to make the new commands resemble the existing ones.

### DIFF
--- a/src/disable.coffee
+++ b/src/disable.coffee
@@ -5,6 +5,7 @@ yargs = require 'yargs'
 
 config = require './apm'
 Command = require './command'
+List = require './list'
 
 module.exports =
 class Disable extends Command
@@ -20,14 +21,23 @@ class Disable extends Command
     """
     options.alias('h', 'help').describe('help', 'Print this usage message')
 
+  getInstalledPackages: (callback) ->
+    options =
+      argv:
+        theme: false
+        bare: true
+
+    lister = new List()
+    lister.listBundledPackages options, (error, core_packages) ->
+      lister.listDevPackages options, (error, dev_packages) ->
+        lister.listUserPackages options, (error, user_packages) ->
+          callback(null, core_packages.concat(dev_packages, user_packages))
+
   run: (options) ->
     {callback} = options
     options = @parseOptions(options.commandArgs)
 
     packageNames = @packageNamesFromArgv(options.argv)
-    if packageNames.length is 0
-      callback("Please specify a package to disable")
-      return
 
     configFilePath = CSON.resolve(path.join(config.getAtomDirectory(), 'config'))
     unless configFilePath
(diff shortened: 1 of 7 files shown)


Fishrock123:
I'm not sure if it is within the domain of this, but would there be a way to use this to make sure disabled packages are not re-enabled on upgrade?

bronson:
I haven't seen that happen...  I've seen a corruption of ~/.atom/config.cson cause it to be replaced with the defaults, re-enabling all the packages, but that was a sharing-config-on-dropbox issue.

Do you have an issue # for it?

mnquintana:
@Fishrock123 That's an unrelated settings-view issue - I can't seem to find the issue # right now though.

mnquintana:
This looks pretty good! I did notice though that it's currently possible to pass in a non-existent package name to both `enable` and `disable` (`disable` will add it to the `core.disabledPackages` list) – would you mind logging an error for any non-existent package names that are passed to these commands? (And adding specs for that?)

src/disable.coffee
```diff
@@ -0,0 +1,56 @@
...
+    result = _.union(disabledPackages, packageNames...)
+    _.setValueForKeyPath(settings, keyPath, result)
+
+    try
+      CSON.writeFileSync(configFilePath, settings)
+    catch error
+      callback "Failed to save `#{configFilePath}`: #{error.message}"
+      return
+
+    process.stdout.write "Disabled: #{packageNames.join(', ')} "
```
mnquintana: Pretty minor, but I think we can just use `console.log` here (I know it's `process.stdout.write` elsewhere in apm), although they're so similar that it doesn't really matter

mnquintana: What do you think about having a newline for each disabled package? I think it would make it a bit more readable if you're disabling a bunch of commands at once, and easier to provide more granular feedback to the user (e.g. if one of the packages was non-existent or something)

src/disable.coffee
```diff
@@ -0,0 +1,56 @@
...
+      return
+
+    try
+      settings = CSON.readFileSync(configFilePath)
+    catch error
+      callback "Failed to load `#{configFilePath}`: #{error.message}"
+      return
+
+    keyPath = '*.core.disabledPackages'
+    disabledPackages = _.valueForKeyPath(settings, keyPath) or []
```
mnquintana: :art: We usually use the existential operator `?` for this

src/disable.coffee
bronson:
```
~/apm (master)⚡$ grep -r stdout.write src | wc -l
      36
~/apm (master)⚡$ grep -r console.log src | wc -l
      41
```

console.log wins by a hair!  Agreed, converting to console.log.

bronson: Now it outputs like this:

~/apm $ bin/apm enable fee fi fo fum aiee
Enabled:
  fee
  fi
  fo
  fum
  aiee 
✓

Better?

src/disable.coffee
bronson: Unless I'm missing something, that won't work.  If I change that line to:

```coffeescript
disabledPackages = _.valueForKeyPath(settings, keyPath)? or []
```
Then the return value ends up either true or false -- the disabledPackages gets thrown away.  Right?

mnquintana: It's been awhile now :sweat_smile: but I think I meant this:

```coffee
disabledPackages = _.valueForKeyPath(settings, keyPath) ? []
```

It's really just a :art: thing since in this case the existential operator works exactly the same as the conditional assignment you have here.

bronson: Ah, gotcha.  Yep, that compiles down to

```js
disabledPackages = (ref = _.valueForKeyPath(settings, keyPath)) != null ? ref : [];
```

Which looks and works great.  Change made.

bronson:
@mnquintana hope I've addressed all your concerns.

It might seem a little odd that the Disable command calls into the List command to retrieve the list of installed packages, but I couldn't find a better way of doing it that didn't duplicate a ton of the List command's code.  One way would be to extract the listing functionality out of the List command and into a shared module but that's a pretty serious change.  This seems the least evil of the alternatives.  :)

New thoughts?

mnquintana:
@bronson Sorry for the super long delay in re-reviewing this - taking a look now

bronson:
@mnquintana how's it look?  :)

src/disable.coffee
```diff
@@ -0,0 +1,83 @@
...
+
+  run: (options) ->
+    {callback} = options
+    options = @parseOptions(options.commandArgs)
+
+    packageNames = @packageNamesFromArgv(options.argv)
+
+    configFilePath = CSON.resolve(path.join(config.getAtomDirectory(), 'config'))
+    unless configFilePath
+      callback("Could not find config.cson. Run Atom first?")
```
mnquintana: It might be a good idea to add a spec for this case (when there is no `config.cson`)

mnquintana: Just kidding, there totally is one already :+1:

lee-dohm:
Looks good to me! :ship:

mnquintana:
Sorry about the delay here @bronson, but thanks so much for adding these! apm just got significantly more powerful :rocket:

bronson:
Yay, thanks!

modulexcite:
Has this been released into production yet?

mnquintana:
@modulexcite It has! It should work with the latest production version of Atom.

modulexcite:
Thanks for your quick response.Would be kind enough to tell me what the command is to list all installed but disabled/enabled packages from apm.

bronson:
I don't think there's a way to do it explicitly.  `grep -v (disabled)` (or some ()-escaped variant) should do it now.

It probably wouldn't be too hard to add an `apm list --disabled` or `apm list --enabled` options.

modulexcite:
Sorry I forgot to mention that I'm on Windows 7 OS and can't do grep.

bronson:
Well then...   I vote for submitting a pull request.  :)

bronson:
Or, heck, I'll submit one myself: #623

`apm list --disabled` is actually pretty handy.

modulexcite:
That would be awesome, thanks.

PR #408, opened 2015-08-03
Update to latest Electron

mnquintana:
Just tested this locally and this should fix the last issue in https://github.com/atom/atom/pull/7877 – apm now builds native modules correctly.

@zcbenz Should `ATOM_NODE_VERSION` and `ATOM_NODE_URL` be renamed to `ATOM_ELECTRON_VERSION` and `ATOM_ELECTRON_URL`, respectively?

/cc @thomasjo @kevinsawicki

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "install": "node ./script/download-node.js",
     "test": "grunt test"
   },
+  "electronVersion": "0.30.2",
   "dependencies": {
     "asar-require": "0.2.0",
     "async": "~0.2.8",
--- a/src/apm.coffee
+++ b/src/apm.coffee
@@ -52,7 +52,7 @@ module.exports =
     process.env.ATOM_REPOS_HOME ? path.join(@getHomeDirectory(), 'github')
 
   getNodeUrl: ->
-    process.env.ATOM_NODE_URL ? 'https://atom.io/download/atom-shell'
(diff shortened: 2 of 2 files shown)


zcbenz:
> @zcbenz Should ATOM_NODE_VERSION and ATOM_NODE_URL be renamed to ATOM_ELECTRON_VERSION and ATOM_ELECTRON_URL, respectively?

Some scripts are still be [hacking the ATOM_NODE_VERSION environment variable](https://github.com/search?q=ATOM_NODE_VERSION&type=Code&utf8=✓), so I think we had better not rename them.

mnquintana:
What about at least renaming the `getNodeVersion` method used internally here?

thomasjo:
I'm in favor of renaming, but since that'd be a breaking change we would have to bump a major version. Since it would probably take a while for everyone to migrate that would mean having to maintain a v1.x development branch for a little while to backport bug fixes etc.

In summary, probably not worth it right now. Changing internal names however shouldn't be a problem, but should go into a separate PR.

thomasjo:
@mnquintana Thanks for helping out by the way! :bow:

kevinsawicki:
> What about at least renaming the getNodeVersion method used internally here?

I think this makes sense, this API isn't exported I believe so it shouldn't break anything.

mnquintana:
Superseded by https://github.com/atom/apm/pull/418

PR #409, opened 2015-08-05
Prepare for Electron

thomasjo:
**NOTE**: This PR essentially supersedes #408. I've cherry-picked the commits from that PR into this one.

In order to finalize https://github.com/atom/atom/pull/7877 we need to bump the version of Node.js used by apm. Prior to this PR that version was hardcoded to v0.10.35. The downloaded version can now be changed via environment variable `APM_NODE_VERSION`, and if not set it currently defaults to v0.10.35 to avoid breaking Atom running on atom-shell.

The approach taken in this PR should allow us to publish a new version of apm without breaking Atom "proper", yet also allow us to change the Node.js version to v0.12.7 in the [Electon PR](https://github.com/atom/atom/pull/7877).

/cc @mnquintana @atom/feedback


thomasjo:
Might be worth changing the default Node.js version to be read from `package.json` similar to `electronVersion`? It would still need to remain v0.10.35 for now regardless.

mnquintana:
> Might be worth changing the default Node.js version to be read from package.json similar to electronVersion?

I like it! :+1: I feel like we should be declaring all of our dependencies in `package.json`, so I think it would be nice to read the Node version from there too.

lee-dohm:
I second the idea :grinning:

script/download-node.js
```diff
@@ -110,7 +115,8 @@ var downloadNode = function(version, done) {
   }
 };
 
-downloadNode('v0.10.35', function(error) {
+var apmNodeVersion = process.env.APM_NODE_VERSION || '0.10.35';
```
kevinsawicki: Just curious why this is needed? This is the version of node that is downloaded with apm, why do multiple versions need to be supported?

src/apm.coffee
```diff
@@ -61,7 +61,7 @@ module.exports =
     process.env.ATOM_API_URL ? 'https://atom.io/api'
 
   getNodeVersion: ->
-    process.env.ATOM_NODE_VERSION ? '0.22.0'
+    process.env.ATOM_NODE_VERSION ? require('package.json').electronVersion
```
kevinsawicki: Should this default to the `electronVersion` in Atom's `package.json` instead? So we don't have to track it in multiple places.

mnquintana: How does https://github.com/atom/apm/commit/dd579c261de158dadbe0d3bb7431404d4ed441f7 look?

script/download-node.js
thomasjo: We have to upgrade Node because of what I think are changes to V8 in conjunction with gyp. Basically we get module mismatches when building native modules. We could just choose to use 0.12.x but that would break Atom on `master`.

Not truly sure what's going on here. There are so many moving parts in play here. All I know is that this approach works against the Atom Electron build, while also not breaking anything.

kevinsawicki: >We have to upgrade Node because of what I think are changes to V8 in conjunction with gyp. Basically we get module mismatches when building native modules. We could just choose to use 0.12.x but that would break Atom on master.

@thomasjo oh okay, thanks for the details.

@zcbenz so is node 0.10 unable to build native modules for Electron 0.30? Should we just start shipping 0.12 instead of 0.10 then?

thomasjo: @kevinsawicki I might be wrong. Things are a bit like a black box for me when it comes to Electron and its bundled Node, and the interaction with apm and its Node.

zcbenz: > @zcbenz so is node 0.10 unable to build native modules for Electron 0.30? Should we just start shipping 0.12 instead of 0.10 then?

Building native modules doesn't rely on the version of Node, both 0.10 and 0.12 are fine.

thomasjo: > Building native modules doesn't rely on the version of Node, both 0.10 and 0.12 are fine.

OK, that's good news. So we might as well simplify my changes and simply specify a single version to download? In other words, no reason to support it being overridden at via environment variables (or similar) at build time?

zcbenz: Node's version has never caused troubles to compiling native modules, so I think we don't need to add a configuration for that.

mnquintana: @zcbenz I'm not sure - I ran into https://github.com/atom/apm/issues/410 by building with an older (0.10.x) of Node.

src/apm.coffee
```diff
@@ -61,7 +61,7 @@ module.exports =
     process.env.ATOM_API_URL ? 'https://atom.io/api'
 
   getNodeVersion: ->
-    process.env.ATOM_NODE_VERSION ? '0.22.0'
+    process.env.ATOM_NODE_VERSION ? require(path.join(@getResourcePath(), 'package.json'))?.electronVersion
```
kevinsawicki: `@getResourcePath` is async so I don't think you'd be able to call it like this.

mnquintana: Derp - sorry about that! :bow: Should be fixed now

thomasjo:
I thiiiiink this one might be good to go (unless we want to cover `getResourcePathSync` with some tests). Would be great if someone has time to help check whether this PR works with an Atom release running on atom-shell. If it does, we should be able to release a new version of apm, and thus resolve the final outstanding issue in https://github.com/atom/atom/pull/7877 :sparkles: 

/cc @atom/feedback

src/apm.coffee
```diff
@@ -48,6 +48,34 @@ module.exports =
           programFilesPath = path.join(process.env.ProgramFiles, 'Atom', 'resources', 'app.asar')
           callback(programFilesPath)
 
+  getResourcePathSync: (callback) ->
```
mnquintana: :+1:

mnquintana: I was thinking of working to make `getNodeVersion` actually work properly with async, but I'm also totally okay with using a synchronous version of `getResourcePath` - whichever way you think would work better.

thomasjo: I was thinking about that too, but this approach seemed to have less of a ripple effect.

thomasjo: I'll :fire: the unused callback argument later.

Zireael07:
What's holding this up? Judging from atom/atom/#8395, it's a priority thing.

thomasjo:
Since this has been superseded, and is no longer necessary, I'm closing it.

Issue #410, opened 2015-08-05
Building with Node 0.12.x makes all apm commands throw errors


mnquintana:
Building apm with Node 0.12.x and running any apm command throws this error:

```
dyld: lazy symbol binding failed: Symbol not found: _node_module_register
  Referenced from: /Users/mnq2/code/apm/node_modules/keytar/build/Release/keytar.node
  Expected in: dynamic lookup

dyld: Symbol not found: _node_module_register
  Referenced from: /Users/mnq2/code/apm/node_modules/keytar/build/Release/keytar.node
  Expected in: dynamic lookup

./bin/apm: line 34: 51737 Trace/BPT trap: 5       /Users/mnquintana/code/apm/bin/node --harmony_collections /Users/mnquintana/code/apm/bin/../lib/cli.js list
```

Building with Node 0.10.x however seems to work as expected.

Refs https://github.com/atom/apm/pull/407

thomasjo:
This works in #409.

steelbrain:
Bump

PR #417, opened 2015-08-16
Support JS package initialization

stevenhauser:
There's a pretty good amount of desire to be able to write packages in JS in addition to CS, especially with the recentish addition of being able to leverage Babel. However the only way to generate a package currently is to generate it in CS. To facilitate JS package development, this PR adds templates for JS in addition to the existing CS templates.

I tried to keep the JS templates as close to the CS templates with minimal syntax changes, but obviously there's a lot of leeway there so let me know if there's anything you'd like to change.

Note that I branched off of #409 due to [this issue](https://github.com/atom/apm/issues/410), so once that PR is in I may need to clean this up a bit. And I'm having problems running the specs for the same reason, so I'm still looking into that but wanted to get this PR out there.

The next step after this would be to [allow JS package generation within the UI](https://github.com/atom/package-generator/issues/14).


stevenhauser:
Since #409 was closed, I rebased this off of latest `master`, though I still get the errors mentioned above in #410 and https://github.com/atom/atom/issues/3184. I've tried trashing `node_modules`, a few different versions of node, reinstalling command line tools, etc, but still no go. Unsure how to get this running now or if there's any interest in this PR. Let me know if there's anything else I can do here.

thomasjo:
Have you tried building with Node v0.10.40? That should work. /cc @atom/feedback

stevenhauser:
Yeah that's the exact version I've got right now.

smashwilson:
I was just about to start looking at sending a PR like this myself. :+1:

Just to add another data point, this branch builds and tests clean for me on OS X 10.10.5 across a few node and npm versions:

 * node v0.10.40, npm 1.4.28
 * node v0.12.7, npm 2.11.3
 * node v4.1.1, npm 2.14.4

A thought on the implementation: rather than adding a `--js` flag, what would you think about making it a `--lang` parameter that could accept a parameter of `coffeescript` or `javascript-es6`, defaulting to `coffeescript` for backwards compatiblity? It feels a little more explicit and future-proofed than `--js` to me.

50Wliu:
> what would you think about making it a --lang parameter that could accept a parameter of coffeescript or javascript-es6, defaulting to coffeescript for backwards compatiblity?

That sounds pretty nice to me :smile:.  :+1:

stevenhauser:
@smashwilson thanks for the additional build testing. I have no idea what's happening on my machine.

I like your `--lang` suggestion, though I think instead of `javascript-es6`, just `javascript` should be good. This prevents the need to maintain JS versions, deal with the fact that it's ES2015 now, etc. I'll work on updating this PR when I've got some time.

smashwilson:
> I have no idea what's happening on my machine.

Yeah, even after reading through the linked issues, I'm not sure either. I'm assuming the failures are independent of this PR. Can you replicate it on master?

> I like your --lang suggestion, though I think instead of javascript-es6, just javascript should be good. This prevents the need to maintain JS versions, deal with the fact that it's ES2015 now, etc.

Reasonable. I could also see `javascript-babel` because it's really using whatever flavor Babel supports, configured as bundled with Atom. That would let us reserve `javascript` for "pure JavaScript, zero transpiling or anything, exactly what you give it in the tin" which some may prefer. (I certainly want the Babel flavor first though :wink:)

.gitignore
```diff
@@ -1,6 +1,6 @@
 node_modules/
 bin/node*
-lib/
+/lib/
```
mnquintana: :art: Why'd you prepend the `/` here?

stevenhauser: Noted in 5477e05

> This is to avoid ignoring the `lib/` directory within packages within `templates/` which requires force adding templates.

So basically to only ignore the top level `lib`.

stevenhauser:
Managed to get this building with `v0.10.40` on a different machine; still have no idea what's busted with the other one. 

While I was about to implement `--lang`, I did remember that there's already a `language` option, so I think `--lang` would be pretty :corn:fusing. Maybe `--syntax`? Anyone have any better ideas?

Regarding `javascript` vs `javascript-babel` @smashwilson, I see your point, but I'd be inclined to wait until someone says they don't want babel before implementing. I'm just hesitant to lock down to a specific ES version, transpiler, etc, especially when I think most devs will want babel, and those who don't aren't prevented from writing ES5. Plus maintaining a third "language" would suck (to support CS + ES2015 + untranspiled JS).

stevenhauser:
Ok, updated this PR:

1. Ended up using `--syntax` over `--lang` because `--language` is already an option and I think that'd be really :corn:fusing.
2. Went with `javascript` and `coffeescript` as possible `--syntax` values. I really think most devs will be fine with transpiled JS, and I didn't want to a) lock down to a JS version name, b) lock down to a transpiler, or c) maintain a third syntax.

Also, for anyone from the future, I ended up fixing my [node problems](https://github.com/atom/apm/pull/417#issuecomment-145371341) by trashing `nvm` and and `~/.npm`, reinstalling it, recloning this repo, doing an `npm install` *with node v0.10.40 as my `nvm` version*, and then everything worked. Probably overkill, but it's all over. I think something with `nvm` may have been the core issue but I'm unsure.

MadLittleMods:
How do I depend on this branch now `stevenhauser/apm#support-js-package-initialization`?
 - Is it as simple as replacing `C:\Users\some-user\AppData\Local\atom\app-1.0.19\resources\app\apm`?

templates/package-javascript/lib/__package-name__-view.js.template
```diff
@@ -0,0 +1,29 @@
...
+
+export default class __PackageName__View {
+
+  constructor(serializedState) {
+    // Create root element
+    this.element = document.createElement('div');
+    this.element.classList.add('__package-name__');
+
+    // Create message element
+    message = document.createElement('div');
```
MadLittleMods: Use `let`/`var`: `let message = document.createElement('div');

stevenhauser:
@MadLittleMods not sure what the "official" way is, but you could just symlink your `apm` to the one on this branch. Not familiar with how that works on Windows.

templates/package-javascript/lib/__package-name__-view.js.template
stevenhauser: Great catch, missed that in the conversion from CS.

templates/package-javascript/lib/__package-name__-view.js.template
```diff
@@ -0,0 +1,29 @@
...
+  // Tear down any state and detach
+  destroy() {
+    this.element.remove();
+  }
+
+  getElement() {
+    return this.element;
+  }
+
+};
```
MadLittleMods: semicolon is not necessary here, see *export default HoistableDeclaration*: https://github.com/eslint/eslint/issues/2194#issuecomment-88575851

Willyham:
:+1: Would love to see this get merged

mertkahyaoglu:
Yaay. Looking forward to seeing this get merged as well.

Other than that, I think you should delete the semicolons at the end of the lines. You can check the language stardart used in [this](https://github.com/atom/line-ending-selector) package.

thedaniel:
At this point I believe @atom/core plans to standardize (lol) on Standard JS style, so it would be good if this matched that.

stevenhauser:
Oh geez not this semicolon debate again. Once a style guide is 100% confirmed, I'll update this to match it.

thedaniel:
Let's not worry about style for now as long as it's valid JS. At some point we'll start moving to JS in earnest and make decisions about core, and update this here. You're right that having that discussion here is a derail.

thedaniel:
@atom/core I'd like to merge this and update the JS template syntax in the future when we make a public decision around a JS styleguide. If there aren't any strong objections by the end of the week I'll hit the big green button.

nathansobo:
No objections here. Styling can always be adjusted.

thedaniel:
dstfp

Excited about this!

stevenhauser:
Awesome, thanks for the merge! Next, I intend to get this into the UI.

flq:
Hi! So, um, I guess I still have to use cs for packages? Or is it closed?

lee-dohm:
Atom has supported using JS in packages for over a year: http://blog.atom.io/2015/02/04/built-in-6to5.html

PR #418, opened 2015-08-17
Use electron version from Atom's package.json

kevinsawicki:
This pulls the `electronVersion` from Atom's `package.json` so the version doesn't need to be bumped  in this project for new versions.

@thomasjo and @mnquintana this is my take on what #409 is trying to do. I am totally onboard with moving the version configuration out of this repository and into atom/atom. I'm cautious to make upgrading to node 0.12 part of this though since I'd like to de-tangle upgrading the node version bundled with apm from upgrading Atom to using the latest Electron.

Refs #409

### DIFF
--- a/src/command.coffee
+++ b/src/command.coffee
@@ -66,3 +66,19 @@ class Command
       version.replace(/-.*$/, '')
     else
       version
+
+  loadInstalledAtomMetadata: (callback) ->
+    @getResourcePath (resourcePath) =>
+      try
+        {version, electronVersion} = require(path.join(resourcePath, 'package.json')) ? {}
+        version = @normalizeVersion(version)
+        @installedAtomVersion = version if semver.valid(version)
+
+      @electronVersion = process.env.ATOM_NODE_VERSION ? electronVersion ? '0.22.0'
+      callback()
+
+  getResourcePath: (callback) ->
+    if @resourcePath
+      process.nextTick => callback(@resourcePath)
+    else
+      config.getResourcePath (@resourcePath) => callback(@resourcePath)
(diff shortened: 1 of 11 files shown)


thomasjo:
This looks brilliant :tophat:

src/command.coffee
```diff
@@ -66,3 +69,19 @@ class Command
...
       version
+
+  loadInstalledAtomMetadata: (callback) ->
+    @getResourcePath (resourcePath) =>
+      try
+        {version, electronVersion} = require(path.join(resourcePath, 'package.json')) ? {}
+        version = @normalizeVersion(version)
+        @installedAtomVersion = version if semver.valid(version)
+
+      @electronVersion = process.env.ATOM_NODE_VERSION ? electronVersion ? '0.22.0'
```
mnquintana: If we want to eventually rename this to `ATOM_ELECTRON_VERSION`, what do you think about starting to support that now by adding a case for `process.env.ATOM_ELECTRON_VERSION` before `process.env.ATOM_NODE_VERSION`? This would make it a little easier to deprecate `ATOM_NODE_VERSION` and remove it for apm 2.0.

kevinsawicki: Great idea, will do :+1:

thomasjo:
Let's :ship: this? I'd love to be able to bump apm in https://github.com/atom/atom/pull/7877 to unblock it.

kevinsawicki:
Published in 1.0.3 :+1:

steelbrain:
I'm on atom `v1.3.2` and I still have a super-ancient node bundled with it (`v0.10.40` to be precise) which causes https://github.com/atom/apm/issues/410

PR #421, opened 2015-08-21
Rig getAtomDirectory to understand Win32 Portable Mode

paulcbetts:
This PR matches up with atom/atom#8442 to allow users to install packages into their portable `.atom` folder

## TODO:

- [ ] Make sure this actually works
- [ ] Code Review


damieng:
Is this still required? Did it get forgotten about?...

PR #425, opened 2015-09-01
:art: beautified dependencies

fscherwi:


kevinsawicki:
Thanks for fixing these up but `^` is only supported in newer versions of `npm` so this old style is used to support a wider range of versions.

PR #426, opened 2015-09-01
improve for older npm versions

fscherwi:


Arcanemagus:
It looks like this PR can be closed now as it seems `apm` has moved past where it was necessary.

mnquintana:
Thanks @Arcanemagus! I'm going to close this out – we don't plan on supporting npm versions that don't support caret semver ranges.

PR #428, opened 2015-09-09
Remove fallback electron version

maxbrunsfeld:
CI was failing because `apm` was building Atom's native modules against the wrong version of Electron. The `ATOM_ELECTRON_VERSION` environment variable was not set, and `/Applications/Atom.app` was not present, so `apm` was silently falling back on version `0.22.0`.

This removes that fallback version, and makes `apm install` fail if it can't find the version of electron to build against.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "atom-package-manager",
   "description": "Atom package manager",
-  "version": "1.0.5",
+  "version": "1.0.6-0",
   "license": "MIT",
   "repository": {
     "type": "git",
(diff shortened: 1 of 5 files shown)


thomasjo:
Nice! LGTM :ship:

bronson:
Can this be closed now that https://github.com/atom/apm/commit/687882d96eea03fd8d727e25708ac31dfd9590f2 is merged?

maxbrunsfeld:
Yeah, I totally forgot about this. Thanks @bronson!

PR #430, opened 2015-09-11
[WIP] Update to Use Latest Node

jlord:
:construction: 

Upgrading `apm` to use the latest Node.js which provides a version for ARM. 

:construction:

### DIFF
--- a/script/download-node.js
+++ b/script/download-node.js
@@ -49,9 +49,11 @@ var downloadTarballAndExtract = function(url, location, callback) {
 };
 
 var copyNodeBinToLocation = function(callback, version, targetFilename, fromDirectory) {
-  var arch = process.arch === 'ia32' ? 'x86' : process.arch;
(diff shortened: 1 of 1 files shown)


jlord:
Currently investigating a couple of modules that aren't compiling on `4.0.0`.

Using some voodoo (printing out each module that was being required to see where it fails) @maxogden helped me find which modules are causing this error when running `bin/apm`:

```bash
bin/apm: line 29: 31034 Segmentation fault      (core dumped) "$binDir/$nodeBin" "$binDir/../lib/cli.js" "$@"
```
It looks like it's `git-utils` and `first-mate`. I've opened issues in them:

- https://github.com/atom/first-mate/issues/58
- https://github.com/atom/git-utils/issues/56

`node-keytar` (also native) seems to be ok.

jlord:
Looks like `git-utils` was updated not long after I had my issues and I just built it successfully and built `apm` without any changes to `first-mate` soon that's cool :sparkles:

PR #444, opened 2015-10-08
Add missing subclasses to .markup and remove .gfm specific styling

burodepeper:
See https://github.com/atom/one-dark-syntax/pull/50 for more details.

This change shouldn't affect any syntax-themes (unless they have really specific scopes) as `language-gfm` already uses the more generic `.markup` classes.

/cc @simurai

### DIFF
--- a/templates/theme/styles/base.less
+++ b/templates/theme/styles/base.less
@@ -259,6 +259,8 @@ atom-text-editor .search-results .marker.current-result .region,
 }
 
 .markup {
+  -webkit-font-smoothing: auto;
+
   &.bold {
     color: @orange;
     font-weight: bold;
@@ -277,6 +279,7 @@ atom-text-editor .search-results .marker.current-result .region,
     font-style: italic;
   }
 
+  &.heading,
   &.heading .punctuation.definition.heading {
     color: @blue;
   }
@@ -285,8 +288,15 @@ atom-text-editor .search-results .marker.current-result .region,
     color: @green;
   }
 
+  &.link {
+    color: @blue;
+  }
+
   &.list {
-    color: @red;
+    .punctuation,
+    .variable {
+      color: @red;
+    }
   }
 
   &.quote {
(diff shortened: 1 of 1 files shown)


templates/theme/styles/base.less
```diff
@@ -277,6 +279,7 @@ atom-text-editor .search-results .marker.current-result .region,
     font-style: italic;
   }
 
+  &.heading,
   &.heading .punctuation.definition.heading {
```
simurai: What's actually an example for `&.heading .punctuation.definition.heading`? I can't find where that gets used. Maybe we can remove it and just have `&.heading`.

burodepeper: I have no idea. Just remove it, and hope nobody notices anything?
From the default languages, the only one with `.heading` is markdown, right?

simurai: When doing a [search](https://github.com/search?q=user%3Aatom+.punctuation.definition.heading&type=Code&utf8=%E2%9C%93) I can only find it this file: https://github.com/atom/language-coffee-script/blob/master/grammars/coffeescript%20(literate).cson#L131. Seems like when Markdown is used in a CoffeeScript file? History doesn't reveal much either.

> Just remove it, and hope nobody notices anything?

Yeah, I would say so. It sounds really specific for a "base" selector.

burodepeper: Is there any spec that gives an example for this? If not, I'd say remove it.

simurai: There is https://manual.macromates.com/en/language_grammars#naming_conventions, but doesn't go deeper than `.markup.heading`.

burodepeper: I'll take the blame if this destroys the world, okay?

simurai: :smile: Will be fine. And we can always add it back if it's really crucial.

Issue #398, opened 2015-07-15
Respect HTTP_PROXY and HTTPS_PROXY Environment Variables


TuckerWhitehouse:
For some reason, apm does not respect if the HTTP_PROXY and HTTPS_PROXY environment variables are set and instead I must hard code them into my .apmrc file.

NPM respects my environment variables and works without issue.

JPvRiel:
just to add to this, on Linux, the env proxy vars are often lowercase, so http_proxy, https_proxy, ftp_proxy and no_proxy in the shell env should be respected.

50Wliu:
The initial issue should be fixed by #447.  @JPvRiel if you're still experiencing problems with lowercase env vars, please create a new issue for that :sparkles:.

PR #447, opened 2015-10-10
make apm look for proxy env variables if not already in the config

rjschu:
as issue #398 requested I've made it so that npm looks for the environment variables HTTPS_PROXY and HTTP_PROXY to use if the proxy is not set within the config file. 

when setting my HTTPS_PROXY variable to http://proxy.com:8080 it is picked up by apm and added to the request as show in below screenshot:
screenshot from 2015-10-10 15 36 55


iax7:
Awesome change! already waiting for this!

thedaniel:
:+1:

PR #449, opened 2015-10-13
Upgrade to Node v4.x

thomasjo:
I reckon it's about time to head for greener pastures — Node v4. Ensuring apm runs on v4 should help @jlord with getting Atom to run on ARM machines. In addition Node v4 has all the performance improvements, ES6 support, bug fixes, etc. that were added to io.js.

I experimented with bumping npm to v3 (3.3.6) but that proved less trivial due to the removal(?) of `peerDependencies`, so that will require more changes, and testing.

Note that I also bumped npm to v2.14.7 (latest stable v2 version as of writing this); I can easily yank that change if someone wants to keep the changeset here as minimal as possible.

/cc @atom/feedback 

---
Supersedes #430
Refs https://github.com/atom/atom/issues/7822

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,9 @@ notifications:
     on_failure: change
 
 node_js:
-  - 0.10
+  - stable
+  - v4.2
+  - v0.10
 
 git:
   depth: 10
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ notifications:
     on_success: never
     on_failure: change
 
+env:
+  global:
+    - CXX=g++-4.8
+
 node_js:
   - stable
   - v4.2
@@ -21,5 +25,8 @@ sudo: false
 
 addons:
   apt:
+    sources:
+      - ubuntu-toolchain-r-test
     packages:
       - libgnome-keyring-dev
+      - g++-4.8
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,7 @@ env:
     - CXX=g++-4.8
 
 node_js:
-  - stable
   - v4.2
-  - v0.10
 
 git:
   depth: 10
--- a/bin/apm
+++ b/bin/apm
@@ -25,4 +25,4 @@ done
 
 binDir=`pwd -P`
 builtin cd "$initialCwd"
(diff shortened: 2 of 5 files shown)


thedaniel:
I'm happy 4.2 because it's their first [LTS](https://nodejs.org/en/blog/release/v4.2.0/) as well, which is something node really needs and we'll benefit from.

thomasjo:
> I'm happy 4.2 because it's their first LTS as well, which is something node really needs and we'll benefit from.

That's a really good point that I completely forgot to mention! Thanks for the heads up :bow:

thomasjo:
I've rebased this after a bunch of experimentation getting the build correct on Travis CI, but everything is working now. Hence I believe this is good to go :rocket:

.travis.yml
```diff
@@ -5,8 +5,14 @@ notifications:
...
 
+env:
+  global:
+    - CXX=g++-4.8
+
 node_js:
-  - 0.10
+  - stable
+  - v4.2
+  - v0.10
```
kevinsawicki: I'm okay removing `v0.10` since we are completely moving past support for it, any concerns there?

abe33:
:clap: Definitely waiting for this as well!

.travis.yml
thomasjo: I mostly left it in for legacy reasons, so no concerns about removing it. Want me to keep `stable`? For the time being `stable` is aliased to `4.2`, so we're building against the same version twice. I don't think stable will follow along with Argon releases (v4.2.x), assuming I understand their LTS system correctly — I think we should stick with LTS for the time being to minimize our own effort keeping this thing running on the "bleeding edge" releases.

kevinsawicki: I think pinning to the version we ship is fine so just `v4.2` and we can just upgrade here when we upgrade versions going forward.

jlord:
Awesome! This will make ARM testing so much smoother :sparkles:

thomasjo:
I guess someone on the core team will need to cut a new release since this is an npm package.

joefitzgerald:
:tada:

kevinsawicki:
:+1:

nathansobo:
Hey, I've reverted this merge due to issues with Atom's build on CI. Not quite sure what's going on there, but Atom needs to build reliably on CI with this version of apm before we proceed.

PR #453, opened 2015-10-16
List Atom version in `apm --version`

### LINKED ISSUE #382: apm --version should also list Atom version
It's often helpful for debugging to know what version of Atom is being used with `apm`.

jebschiefer:
Closes #382

Please let me know if there is a better approach to this (like using `spawn` in `getPythonVersion` or something else).

### DIFF
--- a/src/apm-cli.coffee
+++ b/src/apm-cli.coffee
@@ -98,33 +98,36 @@ printVersions = (args, callback) ->
 
   getPythonVersion (pythonVersion) ->
     git.getGitVersion (gitVersion) ->
-      if args.json
-        versions =
-          apm: apmVersion
-          npm: npmVersion
-          node: nodeVersion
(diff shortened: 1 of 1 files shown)


evjan:
A great starting point!

It fails the build and you added no test for the new functionality. When I run the tests locally on OS X with your change they fail for the same reason as the build. I'm not quite sure why, did you manage to get them running locally @jebschiefer?

jebschiefer:
@evjan Thanks for checking this out. Looks like I was using a different version of node than Travis CI. I will look into this more this evening or tomorrow. It may have to do with the path returned by `config.getResourcePath`. If that's the case then I'll try the `spawn` approach.

damieng:
No activity on this PR which would require merge conflict resolution and specs.

PR #456, opened 2015-10-21
Fix CI For Node v4

joefitzgerald:


thomasjo:
As discussed "offline", this is not necessary, so I'm going to close it for now. We can revert the revert of #449 when we figure out the underlying problem.

PR #458, opened 2015-10-22
skip dirs starting with a dot, fixes #446

### LINKED ISSUE #446: Hide package names that start with `.` in apm list
By default, npm symlinks installed bins generated by a package to `../.bin/`. Rather than changing this behavior (since it might be useful to keep the bins organized with the packages, and `.foo` isn't a valid package name anyway), `apm list` should omit directories with these names.

Related:

- https://github.com/atom/settings-view/pull/569
- https://github.com/atom/settings-view/issues/348

bronson:


lee-dohm:
Could you add a quick spec to prevent against regressions?

bronson:
It has a test in the existing spec to prevent regressions.

If you want it to have its own spec, that would be a bit of copy pasta but I'm happy to do it.

lee-dohm:
Could you extend the existing spec to ensure that directories starting with a dot are not included?

bronson:
It currently ensures .bin is excluded.  Not quite understanding...  you want me to mock other dotted directories too, make sure .bin isn't a fluke?

glen-84:
Any update on this?

thedaniel:
Looks good to me, sorry about the delay.

PR #466, opened 2015-10-31
Merge two seperate .entity sections into one.

mswift42:
:art: Having two seperate .entity sections leads to confusion. 

This way you don't get people adding subsections to the first .entity entry, which then get overridden by the second .entity entry.


izuzak:
Great -- thanks for cleaning that up. :sparkles:

PR #468, opened 2015-11-06
Set process title in APM

steelbrain:
After this we can do `pidof apm` instead of doing `pidof node`

[Review on Reviewable](https://reviewable.io/reviews/atom/apm/468)


kevinsawicki:
:+1: Thanks!

PR #473, opened 2015-12-02
Display Atom version when 'apm --version' is called

arabelle:
Display Atom version when 'apm --version' is called

image


bronson:
Did you see #453?

damieng:
No activity, no specs, closing.

PR #484, opened 2015-12-29
use common npm abbreviations, fixes #435

### LINKED ISSUE #435: Configure apm install aliases (apm i)
- Alias `apm i` to `apm install`

bronson:
Also updated some tests to use the abbreviations where convenient.


src/view.coffee
```diff
@@ -9,7 +9,7 @@ tree = require './tree'
 
 module.exports =
 class View extends Command
-  @commandNames: ['view', 'show']
+  @commandNames: ['view', 'show', 'v']
```
Alhadis: Shouldn't `info` be added as an alias?

Homebrew uses it, and I'm constantly getting `info` mixed up with `show` or `view`...

bronson:
Ha, sure.  Added @Alhadis's suggestion and rebased to master.

bronson:
Alas, it looks like CI on master is broken.  The failure isn't related to this patch.

joefitzgerald:
I would much prefer that new specs are added for the abbreviations, rather than changing the existing command names (e.g. `install` > `i`). This will ensure there are not regressions for the full command names.

bronson:
Heh, maybe no tests needed?
 https://github.com/bronson/apm/commit/94a730e1231c59f890296977027107952a50321a

bronson:
OK, the patch has been updated to just specify the abbreviations, not test them.

Agreed, it would be nice to have specs for the abbreviations.  I don't see a way to do it that looks clean though...  Best I came up with would be to write tests like this:
```cs
describe 'aliases for install', ->
  runs ->
    expect(apm.findCommand(['i'])).toBe Install
```

Which requires the following patch to `apm-cli.coffee`:

```patch
diff --git a/src/apm-cli.coffee b/src/apm-cli.coffee
index fde6865..ff0a647 100644
--- a/src/apm-cli.coffee
+++ b/src/apm-cli.coffee
@@ -154,7 +154,12 @@ getPythonVersion = (callback) ->
         version = version?.trim()
       callback(version)
 
+findCommand = (command) ->
+  commands[command]
+
 module.exports =
+  findCommand: findCommand
+
   run: (args, callback) ->
     config.setupApmRcFile()
     options = parseOptions(args)
@@ -203,7 +208,7 @@ module.exports =
         else
           showHelp(options)
         options.callback()
-      else if Command = commands[command]
+      else if Command = findCommand(command)
         new Command().run(options)
       else
         options.callback("Unrecognized command: #{command}")
```
Can finish doing that if you think it's worth it.

src/view.coffee
bronson: (more than a year later...)

Sounds good, included.

PR #502, opened 2016-01-31
should check to see if package exists before trying to uninstall it

### LINKED ISSUE #395: `apm rm` returns success even if the package wasn't found
I had a typo in the package name when removing a package and apm gave me the green check mark, so I was confused when it wasn't removed.

I can see maybe not returning a failure status to the shell on removing a non-existent package, but it seems like the user feedback should have some sort of indication that you tried to remove a nonexistent package.

rjschu:
hopefully this will close #395 :smile:


joshaber:
Awesome, thanks!

PR #503, opened 2016-02-01
Update documentation on proxy usage

a-h:
Changed the README.md file to suggest using the apm command line tool to set the proxy configuration item rather than manually editing the .apmrc file.

Manually editing the ~/.atom/.apmrc file doesn't work if you run Atom on Windows by downloading the zip, extracting it and running it from a directory. In this case, the ~/.atom/.apmrc file is overwritten every time the apm command is executed. However, running the command line persists the changes successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,8 +56,13 @@ file:
 strict-ssl = false
 ```
 
-If you are using a proxy you can configure `apm` to use it by setting the
-`https-proxy` config in your `~/.atom/.apmrc` file like so:
+If you are using a proxy you can configure `apm` to use it by either:
+
+* Using apm to add the new configuration item:
+
+```apm config set https-proxy https://9.0.2.1:0```
+
+* Adding the `https-proxy` configuration item to the `~/.atom/.apmrc` file:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,8 +56,13 @@ file:
 strict-ssl = false
 ```
 
-If you are using a proxy you can configure `apm` to use it by setting the
-`https-proxy` config in your `~/.atom/.apmrc` file like so:
+If you are using a proxy you can configure `apm` to use it by either:
+
+* Using apm to add the new configuration item:
+
+```apm config set https-proxy https://9.0.2.1:0```
```
lee-dohm: Could you put the triple-backticks on their own lines like the other instances in the file?

lee-dohm:
If the new first option is the recommended way, should we just remove the second (and more difficult) option from the `README` altogether?

a-h:
I've made some changes as per your suggestions.

lee-dohm:
Awesome, thanks! :+1:

Issue #505, opened 2016-02-05
Cannot log in to apm after changing my github urls (origin) from https to key


mark-hahn:
I recently changed all my github repo urls from https to key-based, for extra security. APM then started asking for my account login. I couldn't log in. I tried getting a new token and I still couldn't login. I switched my github repo url back to https and I still can't login. So I am stuck, unable to update my packages.

Atom version 1.5.0-beta
Ubuntu 14.04
apm 1.6.0
npm 2.13.3
node 0.10.40
python 2.7.3
git 1.7.9.5

lee-dohm:
Related: https://discuss.atom.io/t/changing-repo-url-breaks-apm-account-token/25294

thedaniel:
The calls to the atom.io API don't even look at your git repository, so I am not sure what is causing this. Can you paste the output of an unsuccessful attempt at updating a package?

mark-hahn:
> so I am not sure what is causing this

It could be  coincidence that I changed the keys and then had the problem.

```
roo(~/.atom/packages/markdown-scroll-sync). apm publish 1.2.1
Welcome to Atom!

Before you can publish packages, you'll need an API token.

Visit your account page on Atom.io https://atom.io/account,
copy the token and paste it below when prompted.

Press [Enter] to open your account page on Atom.io. 
Token> _f_L57J4qXZyDTtvgHMoWfQ
Saving token to Keychain ✓
Preparing and tagging a new version ✓
Pushing v1.2.1 tag ✓
Publishing markdown-scroll-sync@v1.2.1 Welcome to Atom!

Before you can publish packages, you'll need an API token.

Visit your account page on Atom.io https://atom.io/account,
copy the token and paste it below when prompted.

Press [Enter] to open your account page on Atom.io. 
```

joshaber:
`apm` checks for the `ATOM_ACCESS_TOKEN` environment variable before checking the keychain, and uses it if it exists: https://github.com/atom/apm/blob/e223652263d55c433a71c210a558a31a995ea58d/src/auth.coffee#L20 

So even though you keep updating the token stored in the keychain, `apm` keeps using the one set by `ATOM_ACCESS_TOKEN`. Update the token saved in `ATOM_ACCESS_TOKEN` or unset that environment variable and you should be good to go.

joshaber:
I've opened https://github.com/atom/apm/pull/507 to change the order of token lookup to avoid this in the future.

mark-hahn:
[...]

PR #507, opened 2016-02-08
Check the keychain for the token before falling back to ATOM_ACCESS_TOKEN.

joshaber:
Otherwise auth failures get confusing: we’d keep updating the token
stored in the keychain when the real problem was that `ATOM_ACCESS_TOKEN`
was out of date. See #505.


lee-dohm:
:metal: :tada: :rainbow:

PR #511, opened 2016-02-23
docs: Fix confusing wording

simplyianm:
When I first read this readme, it sounded like the other way around. This clarifies things.

cc @tejasmanohar @yarabarla


tejasmanohar:
strange, I was confused by the old wording before... but now I can't even figure out what's wrong with it hahaha. now i'm curious- someone wanna fill me in?

damieng:
Manually applied this change as it couldn't be merged.

PR #512, opened 2016-02-23
Upgrade grunt-coffeelint to 0.0.15 

damieng:
With a clean checkout I was getting the error:

    Running "coffee:glob_to_multiple" (coffee) task
    >> 41 files created.
    Warning: Task "coffeelint" not found. Use --force to continue.
    Aborted due to warnings.
    npm WARN grunt-coffeelint@0.0.13 requires a peer of coffeelint@^1 but none was installed.

Upgrading the grunt-coffeelint package from 0.0.13 to 0.0.15 made this go away.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "grunt-cli": "~0.1.8",
     "grunt-shell": "~0.2.2",
     "express": "~3.2.3",
-    "grunt-coffeelint": "0.0.13",
+    "grunt-coffeelint": "0.0.15",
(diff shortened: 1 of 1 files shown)


damieng:
Actually with the exception of npm all the packages can be updated to the latest version and the tests still pass (npm has to stick at 2.14.19 as 3 changes things too much right now).

Let me know if you want a PR that brings everything up to date. (The express dependency version we're actually has a security issue)

Issue #514, opened 2016-03-02
APM not working under LTS version of Node.js


krrg:
I recently updated my system node version to 4.2.4.  When I try to use `apm` however, it gives me the following output:

`/usr/share/atom/resources/app/apm/bin/node: bad option: --harmony_collections`

Are there any plans to support newer versions of Node.js?

alice-em:
I have the same issue on Fedora 23.

lee-dohm:
The apm tool packages its own version of Node. Did you by chance try to replace the one that comes packaged with apm?

krrg:
On my system, the `apm` tool was symlinked to /usr/bin/node

alice-em:
How exactly do I change the symlink on apm to point to the right version of node*, instead of pointing to my global node installation?

*I assume the right node is `/usr/share/atom/resources/app/apm/bin/node`

krrg:
@alice-em did you use a copr repository on Fedora to install atom?  Or did you use the RPM provided by atom?

alice-em:
I used the copr repo.  Should I not use that and instead use the atom .rpm?

krrg:
Yeah, looks like this is a problem with the copr repo.  The RPM from atom comes with the correct version of node, whereas the copr repo version just uses the system version.

alice-em:
alrighty! Thanks, that solved it on my end.

krrg:
Closing this ticket, since this turned out to not be a problem with apm, but with an unofficial distribution channel for atom.

lee-dohm:
Thanks for following up @krrg! Glad to hear it is fixed for you :grinning:

Issue #481, opened 2015-12-21
bad option: --harmony_collections


romu70:
Hi,
Fedora 23 here, Atom 1.3.2, as soon as I run "apm" (just this simple command, no package specified), I get the message:

```bad option: --harmony_collections```

Of course, same error when trying to install a plugin for real.

**Comment by koppeex** [comment]
I got same message when I build atom on Windows7.

```
$ script/build
Node: v4.2.4
npm: v2.13.3
Python: v2.7.5
Installing build modules...
Installing apm...
node.exe: bad option: --harmony_collections
```

**Comment by ZombieHippie** [comment]
Same error here on Windows 8.1
```
Node: v5.5.0
npm: v2.13.3
Python: v2.7.9
Installing build modules...
=> Took 5280ms.

Installing apm...
=> Took 984ms.

Deleting old packages...
node.exe: bad option: --harmony_collections

```

**Comment by 50Wliu** [comment]
This is because apm currently requires a very ancient version of Node for reasons documented here: #410. apm bundles its own version of Node, so if you are running into this issue, please make sure you didn't symlink a newer version of Node by accident to apm (the COPR Atom repository incorrectly does this I believe).

@koppeex, @ZombieHippie: did you attempt to alter the version of Node that apm uses?

@romu70: Did you install Atom from the COPR repo?

**Comment by romu70** [comment]
@50Wliu didn't remember. And to be honest, I removed Atom and switched to VsCode.

**Comment by koppeex** [comment]
@50Wliu 
I tried old node version. but I could not build them.

```
$ script/build
Node: v0.10.43
npm: v1.4.29
Python: v2.7.5
Installing build modules...
gypnpm ERR! runas@3.1.1 install: `node-gyp rebuild`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the runas@3.1.1 install script.
npm ERR! This is most likely a problem with the runas package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node-gyp rebuild
npm ERR! You can get their info via:
npm ERR!     npm owner ls runas
npm ERR! There is likely additional logging output above.

npm ERR! System Windows_NT 6.1.7601
npm ERR! command "c:\\Program Files (x86)\\nodejs\\\\node.exe" "c:\\Program Files (x86)\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "--userconfig=c:\\atom\\build\\.npmrc" "install" "--loglevel" "erro
[...]

PR #521, opened 2016-03-07
Remove harmony collections from node options

ZombieHippie:
This subject arises in the following issues:
#449
#481 
#514 

Here are two articles backing up how this is unnecessary.

Harmony flag is no longer needed in Node 4
https://nodejs.org/en/blog/weekly-updates/weekly-update.2015-08-07/

Harmony flag is not needed in node v4
https://iojs.org/en/es6.html

By the way, my build environment breaks as a result of these flags,
```
D:\Users\Cole\Desktop\atom>"script\build" --install-dir ..\atom-install --build-dir ..\atom-build
Node: v4.3.2
npm: v2.13.3
Python: v2.7.9
Installing build modules...
=> Took 5360ms.

Installing apm...
=> Took 1013ms.

Deleting old packages...
node.exe: bad option: --harmony_collections
```

### DIFF
--- a/bin/apm
+++ b/bin/apm
@@ -25,4 +25,4 @@ done
 
 binDir=`pwd -P`
 builtin cd "$initialCwd"
-"$binDir/$nodeBin" --harmony_collections "$binDir/../lib/cli.js" "$@"
+"$binDir/$nodeBin" "$binDir/../lib/cli.js" "$@"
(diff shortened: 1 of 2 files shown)


ZombieHippie:
I have tried multiple different versions of Node to mitigate this issue, but the only version that works is `0.11.x` (0.11.9) which is the one right before `0.12.x` which starts the break thereafter (this is described [under the **No more --harmony flag** heading on iojs.org es6](https://iojs.org/en/es6.html))

50Wliu:
This is because `apm` currently requires a very ancient version of Node for reasons documented here: #410.  apm bundles its own version of Node, so if you are running into this issue, please make sure you didn't symlink a newer version of Node by accident to apm (the COPR Atom repository incorrectly does this I believe).

As we are on Node 0.10.x and not 4+, I'm closing this.

ZombieHippie:
This turned out to be an issue where my node-gyp failed the first time, and I needed to reclone atom and rebuild with the correct builder in my path. Ooops.

PR #526, opened 2016-03-17
Use ATOM_ACCESS_TOKEN On Install If Set

joefitzgerald:
- Fixes https://github.com/atom/ci/issues/39

/cc @lee-dohm @thedaniel

### DIFF
--- a/src/install.coffee
+++ b/src/install.coffee
@@ -212,6 +212,7 @@ class Install extends Command
       url: "#{config.getAtomPackagesUrl()}/#{packageName}"
       json: true
       retries: 4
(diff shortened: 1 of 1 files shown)


thomasjo:
Looks good to me!

joshaber:
I mentioned this in chat, but for posterity: I'm not sure this is the change we need, unless I'm misunderstanding this (which is entirely possible!).

The rate limit message is coming from github.com's API, not atom.io itself. It's atom.io's usage of GitHub's API that's being rate-limited.

It seems possible this is related to a change I made to atom.io earlier this week, so I've reverted it. Lemme know if you still see rate limiting.

joshaber:
(If the above is right, we should also revert https://github.com/atom/ci/pull/40.)

joefitzgerald:
@joshaber 's theory was correct. This is now resolved.

PR #532, opened 2016-03-29
Fix package installation when package exports bins

kuychaco:


src/install.coffee
```diff
@@ -120,6 +120,7 @@ class Install extends Command
         if installGlobally
           commands = []
           children = fs.readdirSync(nodeModulesDirectory)
+            .filter (dir) -> dir isnt ".bin"
```
BinaryMuse: @kuychaco This is the regression; compare with with [this line](https://github.com/atom/apm/pull/458/files#diff-b7c70225b25510628f14f4501e624184R67). Should probably be

```coffeescript
.filter (dir) -> not dir.match /^\./
```

/cc @thedaniel

PR #539, opened 2016-04-04
Do not throw errors if packages do not exist on apm uninstall --hard

ZombieHippie:
I do not understand how it is an error unless both a dev package exists **and** a package exists.

Uninstall hard **should succeed** if the packages do not exist afterwards.


lee-dohm:
Thanks for the ping @50Wliu.

lee-dohm:
Thanks for this!

PR #550, opened 2016-05-02
Make accidental unpublishes harder

BinaryMuse:
Let's avoid another https://discuss.atom.io/t/apm-publish-help/28414

This PR includes two changes:

* If you `apm unpublish my-package` (without a version), apm attempts to make it *explicitly clear* what's about to happen.
* If you simply presses enter at the confirmation prompt, the default behavior is to cancel the unpublish; only "yes" or "y" will actually perform the destructive operation. Use `--force` if you *really know* you want to unpublish.

/cc @lee-dohm


lee-dohm:
Awesome work!

PR #551, opened 2016-05-04
Fix truncated JSON output

tensor5:
This fixes a bug occurring with recent Node.js versions, where the output of `apm list --json` is truncated after 65536 characters.
In new Node.js (version 5 or 6), [console.log becomes asynchronous](https://nodejs.org/api/console.html#console_asynchronous_vs_synchronous_consoles), so with big outputs, like the above example, [the callback](https://github.com/atom/apm/blob/master/src/list.coffee#L170) that exits the process runs before the output completes.

I tested this patch in the [apm that I package for Arch Linux](https://github.com/tensor5/arch-atom), where I use the system installed Node rather than the one that ships with apm, and it seems to work fine.

### DIFF
--- a/src/featured.coffee
+++ b/src/featured.coffee
@@ -62,7 +62,7 @@ class Featured extends Command
       return callback(error) if error?
 
       if options.argv.json
-        console.log(JSON.stringify(packages))
+        process.stdout.write JSON.stringify(packages) + '\n', callback
       else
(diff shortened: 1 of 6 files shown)


tensor5:
No longer necessary after 3b6e3def90acaa65962752dff38b0f4556e9536c.

PR #559, opened 2016-05-29
Fix publish test

freebroccolo:
This PR fixes the test in `publish-spec.coffee` which fails on OSX (10.11.5) due to expecting the wrong error output.

### DIFF
--- a/spec/publish-spec.coffee
+++ b/spec/publish-spec.coffee
@@ -35,7 +35,7 @@ describe 'apm publish', ->
       callback.callCount is 1
 
     runs ->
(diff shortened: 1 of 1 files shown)


spec/publish-spec.coffee
```diff
@@ -35,7 +35,7 @@ describe 'apm publish', ->
       callback.callCount is 1
 
     runs ->
-      expect(callback.mostRecentCall.args[0].message).toBe 'Error parsing package.json file: Unexpected token }'
```
50Wliu: How about replacing `.toBe` with `.toContain`?

PR #562, opened 2016-06-01
:arrow_up: npm@3.9.3

### LINKED ISSUE #322: Unable to install packages behind corporate proxy
I get the following error when trying to install a package with apm behind a corporate proxy. 

```
$ apm install linter
Installing linter to /home/marcuslt/.atom/packages 
gyp info it worked if it ends with ok
gyp info using node-gyp@1.0.2
gyp info using node@0.10.35 | linux | x64
gyp http GET https://atom.io/download/atom-shell/v0.21.0/node-v0.21.0.tar.gz
gyp WARN install got an error, rolling back install
gyp ERR! install error 
gyp ERR! stack Error: tunneling socket could not be established, cause=Parse Error
gyp ERR! stack     at ClientRequest.onError (/opt/atom/resources/app/apm/node_modules/npm/node_modules/request/node_modules/tunnel-agent/index.js:168:17)
gyp ERR! stack     at ClientRequest.g (events.js:180:16)
gyp ERR! stack     at ClientRequest.emit (events.js:95:17)
gyp ERR! stack     at Socket.socketOnData (http.js:1593:9)
gyp ERR! stack     at TCP.onread (net.js:528:27)
gyp ERR! System Linux 3.13.0-46-generic
gyp ERR! command "/opt/atom/resources/app/apm/bin/node" "/opt/atom/resources/app/apm/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "install" "--target=0.21.0" "--dist-url=https://atom.io/download/atom-shell" "--arch=x64" "--ensure" "--proxy=http://wwwproxy:3128/"
gyp ERR! cwd /home/marcuslt/.atom
gyp ERR! node -v v0.10.35
gyp ERR! node-gyp -v v1.0.2
gyp ERR! not ok 

```

All proxy settings should be correct and it worked perfectly in earlier versions.

```
$ apm --version
apm  0.142.0
npm  2.5.1
node 0.10.35
python 2.7.6
git 1.9.1
```

### LINKED ISSUE #436: npm@3
npm 3 is going final next week (https://github.com/npm/npm/releases/tag/v3.3.4) with loooots of improvements for Windows users. Just making this an issue so it is tracked and upgrading to it not forgotten.

### LINKED ISSUE #463: apm requires npm@2 module paths
It'd be really great if `apm` could be assembled with npm@3, since it shortens paths and prevents long paths like `apm/node_modules/npm/node_modules/npm-registry-client/node_modules/concat-stream/node_modules/readable-stream/node_modules/process-nextick-args/package.json` on Windows.

Unfortunately, if you `npm install atom-package-manager; npm dedupe` using npm 3, it flattens the entire dependency tree into a single, top-level `node_modules` folder. This causes APM to throw exceptions at runtime at lines like https://github.com/atom/apm/blob/c6e5123b17c99233b661a616ce4e09c38c2c9a07/src/deprecated-packages.coffee#L1, because it assumes `npm/node_modules/semver` will exist. In npm@3, `npm` and `semver` are placed in the same directory.

kuychaco:
Fixes #463 
Fixes #322
Fixes #436


BinaryMuse:
Fixes https://github.com/atom/apm/issues/463

50Wliu:
As well as #322 and #436.

kuychaco:
@atom/core @damieng @atom/feedback - this is ready for review 👀 , thoughts and feedback welcome!

BinaryMuse:
Okay, this uses a new installation strategy that we came up with with @kevinsawicki. Previously, we installed packages by creating a blank folder with a `node_modules` subfolder, and doing `npm install /path/to/downloaded/package.tgz`, which installed the apm package in the `node_modules` subfolder. We then grabbed this lone package folder and moved it into `~/.atom/packages`.

With the new npm v3 package flattening mechanism, we now *also* get a large number of the package's dependencies at the same level, making it difficult to move the package and all its dependencies into `~/.atom`. It's not enough to just move all the other folders into the package's `node_modules` folder, because npm will only flatten packages if they don't conflict, so if `my_package` depended on `underscore@1` and also depended on `awesome_module` which itself depended on `underscore@2`, we'd get conflicts when performing the move.

Instead, we now unpack the downloaded `.tgz` file from GitHub into a temporary folder and run `npm install` in it directly, then move that whole folder into `~/.atom`. In order to maintain the same behavior (e.g. don't install dev dependencies, don't run prepublish hooks, etc.) we pass some flags to npm and also directly modify the package's `scripts` entries if they have any. A test for this is included.

Additionally, since npm no longer installs the base `.tgz` file directly, it will not cache it — we've disabled the test that checks for this. It will, however, still cache all the files for all the dependencies, which we feel is the important part of this.

This feels pretty heavy handed, but seems to be working, and npm@3 brings some huge wins, so we'd love to get some 💭  around this!

as-cii:
Thanks for taking this on, @BinaryMuse and @kuychaco! 👏 ❤️ 💯 

> Additionally, since npm no longer installs the base .tgz file directly, it will not cache it — we've disabled the test that checks for this.

Can we delete it entirely? We can put it back at some point in the future, in case things change.

> This feels pretty heavy handed, but seems to be working, and npm@3 brings some huge wins, so we'd love to get some 💭 around this, and on the caching decision.

👍 on the caching decision: so long as package dependencies are cached, I think we're good.

I am a bit more concerned about removing lifecycle hook scripts, although I guess there's always a defined set of them and we should be fine as long as we remove them all. 

-------------------------

Thinking more about this, I wonder if we could use a variation of the approach we're taking right now on master: the nice thing about it is that it lets `npm` manage all the complexity of package-related concerns, so that we just perform the copy task.

So, what about doing exactly the same thing we were doing before, but pass the `--global-style` argument to `npm install`? From the docs:

> The --global-style argument will cause npm to install the package into your local node_modules folder with the same layout it uses with the global node_modules folder. Only your direct dependencies will show in node_modules and everything they depend on will be flattened in their node_modules folders. This obviously will elminate some deduping.

What this means is that we can just copy the folder, but still have the deduping happen on a per-package basis (which, if I understand correctly, seems to be what we would be doing with this approach too). Thoughts? 💭

BinaryMuse:
So despite the work put into the diagrams, they're simply not necessary because @as-cii seems to have tracked down the flag we were looking for through all the RTFM-ing we did yesterday. Not sure how we missed it. :) Looks like the flag turns off package flattening for the first-level of `node_modules`, which is precisely what we need.

@kuychaco Take a look at the new diff; if you're happy we can `reset --hard` and apply the two latest commits directly on top of 9d99f9770443334636d6735ca7adf43939f43d61.

Issue #568, opened 2016-06-04
Several (incorrect?) assumptions about node path


gjtorikian:
I'm looking at getting Atom running on ARM. I followed the build instructions in the README, but ran into two problems (which are actually the same problem):

* After cloning and building, I tried to run `bin/apm -v` to make sure things were working. I received an error message that said: `/home/gjtorikian/Development/apm/bin/node: No such file or directory`. I traced this back to [this line](https://github.com/atom/apm/blob/b25e8fee56a94e9a884307e87139e1322ca0d755/bin/apm#L35).
* Following the rest of the instructions, I ran `./bin/npm run build`. This crashed with a similar error [on this line](https://github.com/atom/apm/blob/b25e8fee56a94e9a884307e87139e1322ca0d755/script/check-version.js#L7).

So, in both cases, it looks like it's assuming a `node` binary to be in the `bin` folder. I don't see how that's possible with the current instructions or setup. Is this a documentation error or a configuration error? In any event, I was able to move forward by just changing both paths to use `node`, since it's already on my `PATH`.

NoahVlncrt:
I'm trying to create this for myself and I ran into this issue how exactly did your code look if you don't mind me asking?

Issue #569, opened 2016-06-05
Missing path to `node-gyp`?


gjtorikian:
Similar to https://github.com/atom/apm/issues/568, I had an issue with [the install script](https://github.com/atom/apm/blob/b25e8fee56a94e9a884307e87139e1322ca0d755/src/install.coffee#L29) looking for `node-gyp` in a path that didn't exist. `node-gyp` is not currently a dependency for this repo or `atom/atom`.

In the end, I did `npm install -g node-gyp`, and hardcoded the path to my installed `node-gyp` bin folder.

gjtorikian:
This same issue was in the `dedupe` script, too, fwiw.

BinaryMuse:
There's a postinstall script that should (1) download node and (2) run a dedupe so that node-gyp is moved from npm into the top-level node_modules folder. Is this working now that #568 has been resolved?

gjtorikian:
Sorry this took so long for me to get to. Yes, this is working after #568!

PR #570, opened 2016-06-05
Download the correct Node binary for ARM machines

gjtorikian:
This should fix https://github.com/atom/apm/issues/568.

I noticed for the first time an error message that said:

```
download not found: http://nodejs.org/dist/v4.4.5/node-v4.4.5-linux-arm.tar.gz
```

I checked [the Node distribution page](https://nodejs.org/en/download/current/) and found that the URL for ARM machines includes the ARM CPU version.

### DIFF
--- a/script/download-node.js
+++ b/script/download-node.js
@@ -12,6 +12,14 @@ var getInstallNodeVersion = require('./bundled-node-version')
 
 temp.track();
 
+var identifyArch = function() {
+  var arch = process.arch === 'ia32' ? 'x86' : process.arch;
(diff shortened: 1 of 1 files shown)


gjtorikian:
Looks like this also closes https://github.com/atom/apm/issues/569.

BinaryMuse:
❤️

gjtorikian:
Can I humbly suggest a patch bump so that this can go in alongside the other major apm changes in https://github.com/atom/atom/pull/11897? 🙏

PR #576, opened 2016-06-13
Add arm arch and Node v4.0.0

mikev1963:


50Wliu:
This has already been added.

PR #592, opened 2016-07-16
Fix Windows build

as-cii:
This should unblock using AppVeyor on atom/atom#12170.

The main issue with the previous scripts was that we were building the URL for downloading Node x86 for Windows in a wrong way, which was preventing it from being downloaded. As part of fixing that, we have also restructured the scripts a little bit to make them more similar to their Mac/Linux counterparts.

Please, note that we have disabled running tests because some of them appear to be written in a platform-specific way. I think fixing them can be deferred to another pull-request, so that we can start using the new version of apm on Atom right away. /cc: @damieng 

/cc: @nathansobo @atom/core

### DIFF
--- a/bin/npm.cmd
+++ b/bin/npm.cmd
@@ -1,17 +1,6 @@
 @echo off
 setlocal EnableDelayedExpansion
 
-:: Try to find git.exe in path
-for /f "tokens=*" %%G in ('where git') do set "apm_git_path=%%~dpG"
-if not defined apm_git_path (
-  :: Try to find git.exe in GitHub Desktop, oldest first so we end with newest
-  for /f "tokens=*" %%d in ('dir /b /s /a:d /od "%LOCALAPPDATA%\GitHub\PortableGit*"') do (
-    if exist %%d\cmd\git.exe set apm_git_path=%%d\cmd
-  )
-  :: Found one, add it to the path
-  if defined apm_git_path set "Path=!apm_git_path!;%PATH%"
-)
-
 set "PATH=%~dp0;%PATH%"
 set maybe_node_gyp_path=%~dp0\..\node_modules\node-gyp\bin\node-gyp.js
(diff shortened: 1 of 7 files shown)


as-cii:
Before merging this, I am going to publish a pre-release version so that we can test it on Atom.

as-cii:
After upgrading the dependency on Atom, I confirm [this makes the build green](https://ci.appveyor.com/project/Atom/atom/build/1984). 💚

bin/npm.cmd
```diff
@@ -1,17 +1,6 @@
 @echo off
 setlocal EnableDelayedExpansion
 
-:: Try to find git.exe in path
```
damieng: Why is this removed?  It was added so that people on Windows using APM on the command line installing packages that look for git.exe could actually find the one installed by Git Desktop because it isn't in the path...

It shouldn't be interfering with AppVeyor builds in any way.

nathansobo: We didn't understand why it was needed as part of `npm.cmd`... Made more sense in `apm.cmd`. Assumed it was copied and pasted. Am I wrong?

damieng: Are there not scenarios where atom or a user might fire of this npm.cmd
without going via apm.cmd?

as-cii: > Are there not scenarios where atom or a user might fire of this npm.cmd
> without going via apm.cmd?

As far as Atom is concerned I am pretty sure we never invoke `npm` directly. I think the same holds true for users, because they'll interface with `npm` transparently via `apm`. Maybe I am missing something, though? Did you have anything specific in mind, @damieng?

as-cii:
I'll proceed with merging and publishing this. We can refine it later if there's any concern before we roll the railcars on atom/atom.

Thanks @nathansobo and @damieng!

nathansobo:
⚡️

PR #595, opened 2016-07-21
Disable progress when running npm

as-cii:
This regressed in #562, causing every `apm` command that delegates to `npm` to show something like the following:

progress-before

With this pull-request we disable progress report by default in `~/.atom/.apm/.apmrc`, so that running `npm` commands doesn't show the native progress-bar.

progress-after

/cc: @BinaryMuse @kuychaco @nathansobo @atom/core

### DIFF
--- a/src/apm.coffee
+++ b/src/apm.coffee
@@ -112,5 +112,5 @@ module.exports =
         ;
         ; You should instead edit your .apmrc config located in ~/.atom/.apmrc
         cache = #{@getCacheDirectory()}
-
+        progress = false
       """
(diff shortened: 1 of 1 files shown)


src/apm.coffee
```diff
@@ -112,5 +112,5 @@ module.exports =
         ;
         ; You should instead edit your .apmrc config located in ~/.atom/.apmrc
         cache = #{@getCacheDirectory()}
-
+        progress = false
```
thomasjo: Maybe leave a small comment documenting why progress output is disabled?

thomasjo:
I think it would be valuable to document the reason for disabling progress output, but other than that, this looks like a no-brainer 0151 — :ship: it

as-cii:
I have added a comment within `.apmrc` explaining why we're disabling progress report. I was unclear on whether we should have just added it to `apm.coffee` but ultimately this could be less confusing for users that want to write their own `~/.atom/.apmrc`.

I'll go ahead and 🚢 this as soon as the build goes 💚.

PR #536, opened 2016-03-31
fixes #531

hedefalk:
Looking at https://github.com/mapbox/node-pre-gyp/pull/187/commits/ef79b3f52b08469105cd808af6d5a874c9ae5db2 (which doesn't work for apm) I got the feeling that just stating runtime=electron would be a pretty good idea. And it was.

Tried it by simply running:

> 
"/Applications/Atom Beta.app/Contents/Resources/app/apm/bin/node" "/Applications/Atom Beta.app/Contents/Resources/app/apm/node_modules/npm/bin/npm-cli.js" "--globalconfig" "/Users/viktor/.atom/.apm/.apmrc" "--userconfig" "/Users/viktor/.atom/.apmrc" "rebuild" "--target=0.34.5" "--arch=x64" "--runtime=electron"

Without --runtime=electron, node-pre-gyp fails like

* https://github.com/mapbox/node-pre-gyp/issues/188
* https://github.com/strongloop/fsevents/issues/93#issuecomment-146343196
* https://github.com/strongloop/fsevents/issues/93#issuecomment-156270459

PR #596, opened 2016-07-22
Fix missing and hardcoded flags

as-cii:
This pull-request fixes a problem when installing dependencies via `apm` that prevents the runtime from being recognized, causing packages such as nodegit to fail the build. In particular, we are adding the `--runtime=electron` flag to every `npm` command that interfaces with modules installation and their subsequent compilation. 

Please, note that this is the approach that [electron-rebuild uses too](https://github.com/electron/electron-rebuild/blob/27158b165715731836425369c5c5e680420c7cd9/src/main.js#L122-L126).

While in the process of adding such flag, we spotted a couple of places in the `dedupe` command where the target architecture of native modules was being hardcoded; this was probably not an issue, but it feels reasonable to use `config.getElectronArch()` as we do elsewhere.

@atom/core: I'd ❤️ to have some more 👀 on this, so that we can 🚢 it before getting the new `apm` version into beta.

### DIFF
--- a/src/dedupe.coffee
+++ b/src/dedupe.coffee
@@ -36,7 +36,7 @@ class Dedupe extends Command
     installNodeArgs.push("--runtime=electron")
     installNodeArgs.push("--target=#{@electronVersion}")
     installNodeArgs.push("--dist-url=#{config.getElectronUrl()}")
-    installNodeArgs.push('--arch=ia32')
+    installNodeArgs.push("--arch=#{config.getElectronArch()}")
     installNodeArgs.push('--ensure')
 
     env = _.extend({}, process.env, HOME: @atomNodeDirectory)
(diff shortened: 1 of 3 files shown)


50Wliu:
Supersedes #536.

src/dedupe.coffee
```diff
@@ -71,6 +72,7 @@ class Dedupe extends Command
 
   forkDedupeCommand: (options, callback) ->
     dedupeArgs = ['--globalconfig', config.getGlobalConfigPath(), '--userconfig', config.getUserConfigPath(), 'dedupe']
+    dedupeArgs.push("--runtime=electron")
```
hedefalk: Contrary to the name "dedupe", don't we have a lot of dups going on here? I guess we realized we needed '--runtime=electron' on all npm calls and then you had to find 6 places to add it… I found only one in #536 :)

Cleanup time?

maxbrunsfeld: I believe the variable is called `dedupeArgs` because it is the list of arguments passed to `npm dedupe`

maxbrunsfeld: Ah nevermind, didn't get that this was a joke about code duplication 😄

as-cii: That's a great point, @hedefalk. I considered using the `.apmrc` file to store all these configuration settings, but it felt somewhat inconsistent because we still need to use dynamically generated options (e.g. the architecture and the electron version) and there is something nice about having those options close to one another.

I guess an alternative could be to extract a method (and put it in `config.coffee`) named `getNpmBuildOptions()`/`getNpmBuildFlags()` which takes care of constructing an array with the necessary parameters. I'll open a pull-request with this change. 👍

PR #597, opened 2016-07-25
Extract Command.prototype.getNpmBuildFlags()

as-cii:
This pull-request refactors all the methods shelling out to `npm` so that they use `getNpmBuildFlags`, a function that constructs an array of flags useful to build native modules against the Electron version Atom ships with.


bronson:
@as-cii you have a commit bit...  why don'tcha merge this?  :)

smashwilson:
HULK MERGE

PR #615, opened 2016-10-08
Stop using shadow DOM selectors

simurai:
We are in the process of removing the shadow DOM boundary from `atom-text-editor` elements. This pull request upgrades existing selectors so that they:

* Stop using `:host`.
* Stop using `atom-text-editor::shadow`.
* Prepend syntax class names with `syntax--`.

## TODO

- [x] Theme generator (`templates/theme`)
- [x] Remove `:host` from TextMate converter (`src/text-mate-theme.coffee`)
- [x] :arrow_up: first-mate to prepend selectors with `syntax--` when converting TextMate themes (https://github.com/atom/first-mate/pull/80)

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "asar-require": "0.3.0",
     "async": "~0.2.8",
     "colors": "~0.6.1",
-    "first-mate": "^5.0",
+    "first-mate": "6.2.0",
     "fs-plus": "2.x",
     "git-utils": "^4.0",
     "hosted-git-info": "^2.1.4",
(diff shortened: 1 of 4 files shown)


simurai:
@as-cii I removed `:host` from the TextMate theme converter, but I guess we still need to add the `syntax--` prefix. Is it [somewhere here](https://github.com/atom/apm/blob/master/src/text-mate-theme.coffee#L139-L145)?

as-cii:
> but I guess we still need to add the syntax-- prefix.

Great catch, @simurai! ✨ I have updated this pull request's description with a link of the first-mate pull request we'll need to merge in order to address this. Thanks!

as-cii:
Tests are failing, but this seems to be the case on master as well. I'll go ahead and 🚢 this, as it's blocking the shadow DOM removal pull request. Thanks, @simurai

Issue #616, opened 2016-10-11
Disabled packages no longer labelled as "(disabled)" by `apm ls`


michaelsanford:
While debugging a hang on Atom startup (in both Stable and Beta, Windows 10 x64), I notice that I no longer see any distinguishing label on _installed but disabled packages_ from `apm list`.

I see that [list should show "(disabled)"](https://github.com/atom/apm/search?utf8=%E2%9C%93&q=list+disabled) on installed but disabled packages?

Currently, I see a tree representation of all core and community packages which are installed on my system regardless of their enabled state.

Otherwise, `apm` behaves normally. Notably, `apm enable` / `disable` (re #358 #400) _do_ correctly move packages to and from the `disabledPackages` array in `config.cson`.

(I have not filed a bug for the hang-on-start issue, but have not found a resolution. I suppose it's not impossible that they are related.)

bronson:
Some sort of character encoding issue?  Line endings being included in the disabled package names?

Can you put some debug prints on https://github.com/atom/apm/blob/master/src/list.coffee#L48 to see why the package name isn't matching anything in the disabled packages array?

bronson:
In fact, on apm 1.12.5, I think I'm seeing the same thing.  I have a package disabled, but it's not displaying as `(disabled)` in apm list.

Smelling like a bug.

bronson:
Yep, this line: https://github.com/atom/apm/blob/d7fe2b70f4d43f1869fa1a66c3152f235a5570d8/src/list.coffee#L22

Apparently that needs to go through '*' too: `CSON.readFileSync(configPath)?['*']?.core?.disabledPackages`

And FFS, catch-everything clause needs to be fixed.

bronson:
@50Wliu this one should probably be labelled as a bug too.

PR #617, opened 2016-10-14
fix apm list to show packages as disabled again

bronson:
apm doesn't show `(disabled)` anymore after disabled packages.  This PR fixes that.

It looks like this functionality has been broken for a while.

Found by @michaelsanford in https://github.com/atom/apm/issues/616

### DIFF
--- a/spec/list-spec.coffee
+++ b/spec/list-spec.coffee
@@ -3,6 +3,7 @@ fs = require 'fs-plus'
 temp = require 'temp'
 wrench = require 'wrench'
 apm = require '../lib/apm-cli'
+CSON = require 'season'
 
 listPackages = (args, doneCallback) ->
   callback = jasmine.createSpy('callback')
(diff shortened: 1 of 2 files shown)


bronson:
Looks like CI on master is broken?  This branch isn't causing the failure.

bronson:
rebased to master.

this seems like a somewhat important bugfix?

michaelsanford:
I discovered it while debugging an Atom that refused even to start (window got built and nothing happened after that).

Not being able to isolate a package that might be the culprit forced me to (at least temporarily) abandon Atom altogether.

bronson:
@michaelsanford, did you try `atom --safe`?  Just curious.

michaelsanford:
@bronson Fair question; I didn't (probably should have).

I was using it at work and only had a short time I could not be actively developing. As an aside, I'm more accustomed to WebStorm anyway, so I just switched back to that.

Once I figure out the problem, Atom will quickly go back to filling the space between IDE and `sed`.

bronson:
Hi, any comments on this bugfix?

bronson:
Been a month and 4 releases...  Any chance of getting this bugfix in?

bronson:
apm still isn't showing `(disabled)` after package names...  anything to be done here?

50Wliu:
I've brought it up to the team.

PR #638, opened 2016-11-03
Ensure that native modules are built from source

lgeiger:
This PR is build on top of #597 from @as-cii which makes this change much easier 👍 

Native modules relying on prebuilt binaries via [prebuild](https://github.com/mafintosh/prebuild) or [node-pre-gyp](https://github.com/mapbox/node-pre-gyp) have to be [build from source against Electron headers](https://github.com/electron/electron/blob/master/docs/tutorial/using-native-node-modules.md#modules-that-rely-on-node-pre-gyp).

The `--build-from-source` flag ensures that they are built from source otherwise they wouldn't work with Atom.

/cc @rgbkrk

### DIFF
--- a/src/command.coffee
+++ b/src/command.coffee
@@ -111,7 +111,7 @@ class Command
       "--msvs_version=#{vsVersion}"
 
   getNpmBuildFlags: ->
-    ["--runtime=electron", "--target=#{@electronVersion}", "--dist-url=#{config.getElectronUrl()}", "--arch=#{config.getElectronArch()}"]
+    ["--runtime=electron", "--target=#{@electronVersion}", "--dist-url=#{config.getElectronUrl()}", "--arch=#{config.getElectronArch()}", "--build-from-source"]
 
   updateWindowsEnv: (env) ->
     env.USERPROFILE = env.HOME
--- a/src/command.coffee
+++ b/src/command.coffee
@@ -110,6 +110,9 @@ class Command
     if vsVersion = config.getInstalledVisualStudioFlag()
       "--msvs_version=#{vsVersion}"
 
+  getNpmBuildFlags: ->
+    ["--runtime=electron", "--target=#{@electronVersion}", "--dist-url=#{config.getElectronUrl()}", "--arch=#{config.getElectronArch()}"]
+
   updateWindowsEnv: (env) ->
     env.USERPROFILE = env.HOME
 
--- a/src/dedupe.coffee
+++ b/src/dedupe.coffee
@@ -33,10 +33,7 @@ class Dedupe extends Command
 
   installNode: (callback) ->
     installNodeArgs = ['install']
-    installNodeArgs.push("--runtime=electron")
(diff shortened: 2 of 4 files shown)


rgbkrk:
👍

lgeiger:
@50Wliu Please could someone from the @atom team review this PR?

We found a [workaround](https://github.com/nteract/hydrogen/pull/476/files/be2bf68e2cefd88e5d1a1bd1c12990c5a482ea32#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R33) for getting `apm install` to work. But `apm rebuild` still doesn't. Users would need to run `apm rebuild -- --build-from-source` which is a real deal breaker.

lgeiger:
Hi everyone.
We are using prebuilt modules in [Hydrogen for Atom](https://github.com/nteract/hydrogen/) and can't find a way to get `apm rebuild` to work without this PR.

We would really appreciate if you could take a look.

as-cii:
Thanks for the pull request, @lgeiger! 💯 

I am a little unclear on whether we want to enable this by default, as some modules might provide an electron prebuilt binary. In particular, big dependencies like https://github.com/nodegit/nodegit take a significant amount of time to install, so it would be cool to use the prebuilt library whenever possible and fall back to `--build-from-source` only when such binary cannot be found.

Can you elaborate on why exactly node-pre-gyp and prebuild aren't doing what I described above? After a cursory search, it seems like they provide a way of detecting whether a binary is available and [fall back to build](https://github.com/mapbox/node-pre-gyp#options) in case it isn't. Could we pass the `--fallback-to-build` argument instead?

Thanks! 🙇

rgbkrk:
> some modules might provide an electron prebuilt binary

prebuild does not currently provide a way to prebuild binaries for electron: https://github.com/mafintosh/prebuild/issues/61

rgbkrk:
If [`zeromq`](https://github.com/zeromq/zeromq.js) was generally available as a module in Electron it would save us a lot of pain in [Hydrogen](https://github.com/nteract/hydrogen) and [the nteract notebook](https://github.com/nteract/nteract).

lgeiger:
[Electron and Node.js ABIs doesn't match since Electron 1.2.0](https://github.com/electron/electron/issues/5851). This means we cannot use prebuilt binaries which are built against Node.js headers.

This is the reason why it is [recommended to build native modules](http://electron.atom.io/docs/tutorial/using-native-node-modules/#modules-that-rely-on-node-pre-gyp) from source even if `node-pre-gyp` is used.

lgeiger:
I go ahead and close this PR.

I made a [PR](https://github.com/mafintosh/prebuild/pull/135) to [`prebuild`](https://github.com/mafintosh/prebuild/) which ensures that native modules are built from source if the `--runtime=electron` flag is used.

This keeps it possible to support prebuilt binaries for Electron in the future.

as-cii:
Awesome, @lgeiger! Thanks! ✨

PR #639, opened 2016-11-03
Determine remote from package.json

### LINKED ISSUE #343: APM publish uses 'origin' remote from git, rather than repo from package.json
`apm publish X` Will add new tag to origin remote, even if that isn't the correct repo.

The publish process will then fail with 'tag not found' error as the tag doesn't exist in the repo referenced in package.json.

lgeiger:
I often have repositories where the "official" remote isn't `origin`. Most of the time the desired remote would be named `upstream` or similar.

To publish a new version one needs to make sure that `origin` points to the upstream repo where the package should be publish. This behavior can sometimes be very anoying.

This PR will determine the remote from the `package.json` file and fall back to `origin`.
Fixes #343

### DIFF
--- a/spec/packages-spec.coffee
+++ b/spec/packages-spec.coffee
@@ -1,0 +1,17 @@
+Packages = require '../lib/packages'
+
+describe 'getRemote', ->
+  it 'returns origin if remote could not be determined', ->
+    expect(Packages.getRemote()).toEqual('origin')
+
+  it 'returns repository.url', ->
+    pack =
+      repository:
+        type: 'git'
+        url: 'https://github.com/atom/apm.git'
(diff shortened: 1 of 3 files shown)


lgeiger:
I'm using this branch right now to publish all of my atom packages.
This helps a lot if a package is maintained by multiple people where `origin` is often not the desired remote to publish to.

It would be great if this PR could be reviewed.

lee-dohm:
This needs tests in order to be considered for merging.

lgeiger:
@lee-dohm Thanks for taking a look. I moved the function to `src/packages.coffee` and added tests.

smashwilson:
:+1: I like it. Thanks!

lgeiger:
Thanks for merging 👍

PR #645, opened 2016-11-12
Fix "Installing modules failed" on Atom

jsolisu:
I have **winreg** on **package.json** (Atom) linked to **github:jsolisu/node-winreg**, when i use: _script\build_ or _script\bootstrap_ APM fails with _"Installing modoles failed..."_

The problem is because PATH is undefined. I revert changes from 1.15.0 using 1.14.1 version of _command.coffee_

[npm-debug.log.txt](https://github.com/atom/apm/files/587501/npm-debug.log.txt)

### DIFF
--- a/src/command.coffee
+++ b/src/command.coffee
@@ -125,10 +125,11 @@ class Command
 
   addNodeBinToEnv: (env) ->
     nodeBinFolder = path.resolve(__dirname, '..', 'bin')
-    if env['PATH']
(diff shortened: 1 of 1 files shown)


damieng:
I'm curious as to why this is breaking on your system.

Environment variables are case-insensitive in node since v4 https://github.com/nodejs/node/issues/9157

jsolisu:
@damieng It is right. Please note the effect on the image. It´s similar to: https://github.com/atom/atom/issues/11302

error

PR #653, opened 2016-11-29
Update Electron headers URL

kevinsawicki:
Both URLs are still available and serve the same assets, but this one uses the new name for Electron, not the old one Atom Shell.


joefitzgerald:
LGTM

PR #656, opened 2016-12-08
Use https://atom.io/download/electron

50Wliu:
Should be a simple and harmless change.


Ben3eeE:
https://github.com/atom/apm/pull/653?

50Wliu:
Nice!

PR #677, opened 2017-01-26
Replacing async.map with async.mapLimit to adress #641

### LINKED ISSUE #641: `apm outdated` and `apm upgrade` don't work for me; other commands work fine
`apm outdated` and `apm upgrade` are returning with a ETIMEDOUT error.

I initially thought the issue may have been a network issue on my part, but `apm search <package>` and `apm install <package>` work just fine.  For now, I'm upgrading packages by using `apm install <package>`

**Reproduction Steps:**

Run `apm outdated` or `apm upgrade <optional-package-name>`

**Expected behavior:**

List of outdated packages or begin upgrading outdated packages.

**Observed behavior:**

After no output for several seconds:

  * `apm outdated` shows error `Request for package information failed: connect ETIMEDOUT 54.243.164.241:443`
  * `apm upgrade` shows error `Request for package information failed: read ETIMEDOUT`

**Versions (reported by apm):**
  * apm  1.12.5
  * npm  3.10.5
  * node 4.4.5
  * python 2.7.11
  * git 2.8.4

**Atom version:** 1.11.2
**OS and version:** macOS 10.12

**Installed packages:**  will include if necessary on separate post

**Additional information:**

* Problem can be reliably reproduced, doesn't happen randomly: Has been reliably reproduced for last couple of days

bennyborn:
### Description of the Change

Using `async.map` to get the latest versions for all packages probably fires hundreds of simultaneous requests to the webserver (depending on how many packages are installed). Replacing this with `async.mapLimit` ensures that only (in this case) 10 connections are made at once.

### Alternate Designs

No alternates came in mind since async was already the right tool for the job.

### Benefits

Preventing the client from having to open too many simultaneous connections opened which caused real problems for some users which resulting in updates not working at all.

### Possible Drawbacks

Theoretically it could result in a slower update process since we're only requesting information for a maximum of 10 packages at once. Seeing the current problem of people not be able to update at all this might be a drawback they'll likely accept.

### Applicable Issues

Fixes #641

### DIFF
--- a/src/upgrade.coffee
+++ b/src/upgrade.coffee
@@ -129,7 +129,7 @@ class Upgrade extends Command
         @getLatestVersion pack, (err, latestVersion) ->
           done(err, {pack, latestVersion})
 
-    async.map packages, getLatestVersionOrSha, (error, updates) ->
+    async.mapLimit packages, 10, getLatestVersionOrSha, (error, updates) ->
       return callback(error) if error?
 
       updates = _.filter updates, (update) -> update.latestVersion? or update.sha?


50Wliu:
@bennyborn Have you tested these changes on your machine, and does it fix the issue?

bennyborn:
Yep, tested and fixed the problem.

sbrl:
Awesome! When will this make it's way into atom stable, @50Wliu?

50Wliu:
I don't have permission to publish a new version of apm, bit I'll try to get it into 1.15.0.

Issue #670, opened 2017-01-09
apm saying it can’t find existing tag


nwhetsell:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

When attempting to publish a new version of [language-csound](https://atom.io/packages/language-csound), I was bitten by #555 and #558 and eventually resorted to a reset, and then a force push. Things seemed to be working, but I’ve tried to publish another update with `apm publish patch` with the result:

```
Preparing and tagging a new version ✓
Pushing v0.0.28 tag ✓
Publishing language-csound@v0.0.28 ✗
Creating new version failed: Git tag not found
```

This is strange, because the tag exists: https://github.com/nwhetsell/language-csound/releases/tag/v0.0.28. The [package webpage on atom.io](https://atom.io/packages/language-csound) still says the latest version is 0.0.27, not 0.0.28.

I’m not sure what to do next.

### Versions

macOS Sierra v10.12.2 (16C67)

`atom --version` outputs

```
Atom    : 1.12.9
Electron: 1.3.13
Chrome  : 52.0.2743.82
Node    : 6.5.0
```

`apm --version` outputs

```
apm  1.12.9
npm  3.10.5
node 4.4.5
python 2.7.10
git 2.10.1
```

nwhetsell:
This issues appears to have resolved itself.

Issue #558, opened 2016-05-29
Make publish atomic


bali182:
My regular workflow for publishing a new version of a package is the following:

1. Do the work on a feature branch
2. Push it
3. Merge to master
4. Publish it using apm

Now it occurred more than once, that I forgot to pull from master before doing an `apm publish <version>`, to be more general, my local and remote repo was out of sync. It's an honest mistake, I think it can happen to anyone. The problem is, that while apm notifies you about this, a "zombie" version will still be published, that is neither un-publishable, nor re-publishable, it's a persistent reminder of this mistake.

My question is, **would it be possible to make apm publish atomic**? Meaning that it either **succeeds**, or **leaves no trace** in the apm registry (or whatever this data is stored in)?

50Wliu:
:+1:.  I've ran into this many times as well.  To recover you can use the following steps:
1. `git reset HEAD~1 --hard` to erase the `Prepare vx.y.z` commit
2. `git tag -d <version>` to remove the zombie tag
3. If necessary `git push origin :/refs/tags/<version>` to remove the tag from the remote

Then `git pull` and try again.

Issue #459, opened 2015-10-22
apm publish --amend?


danielbayley:
Something like `apm publish --amend` might be useful for really small fixes perhaps? Especially since bumping the version number & pushing tags is so error prone at the moment (see #315 for example…)

thomasjo:
I don't really see a valid use case for this. What's wrong with pushing a patch release? If you made a big mistake, unpublish and republish? If something failed during a publish, you can try republishing the tag via `apm publish --tag <tag>`.

You're claiming that bumping version numbers and pushing tags is something that is error prone, how would this not suffer from the same problems? How would this proposed `--amend` flag work?

Thanks in advance!

danielbayley:
Ah, I wasn't aware that `apm publish --tag <tag>` would basically do this (i.e attempt the publish again without an additional version bump or adding another commit…) I guess `--amend` could just be an alias for `apm publish --tag <latest tag>` in that case… or not?

thedaniel:
That's what patch releases are for. You shouldn't be modifying already-published releases of any software because that undermines the idea of having a known stable configuration.

danielbayley:
Fair enough, I get that.

@thedaniel I only really suggested this after borking the `test` and `language-jxa` packages though (deleted the GitHub repo before `apm unpublish`), which I would *really* like deleted from the database if possible (see #221).

Issue #422, opened 2015-08-26
apm publish failed


nobuhito:
```
macpro-3:atom-syntax-changer nobu$ apm --version
apm  1.0.1
npm  2.5.1
node 0.10.35
python 2.7.6
git 1.8.4
macpro-3:atom-syntax-changer nobu$ cat package.json 
{
  "name": "atom-syntax-changer",
  "main": "./lib/syntax-changer",
  "version": "0.0.1",
  "description": "Syntax theme change by buffer grammar.",
  "keywords": [
    "syntax",
    "gramar",
    "theme"
  ],
  "repository": "https://github.com/nobuhito/atom-syntax-changer",
  "license": "MIT",
  "engines": {
    "atom": ">=1.0.0 <2.0.0"
  },
  "dependencies": {
    "atom-space-pen-views": "^2.1.0"
  },
  "consumedServices": {
    "status-bar": {
      "versions": {
        "^1.0.0": "consumeStatusBar"
      }
    }
  }
}
macpro-3:atom-syntax-changer nobu$ apm publish patch
Preparing and tagging a new version ✓
Pushing v0.0.2 tag ✓
Publishing atom-syntax-changer@v0.0.2 ✗
Creating new version failed: Git tag not found
```

same as https://atom.io/packages/syntax-changer

PR #685, opened 2017-02-09
Wait 10s for the tag to be found on `apm publish`

JoshCheek:
### Description of the Change

Wait 10s for the tag to be found on `apm publish`. I think there are situations where [5s](https://github.com/atom/apm/blob/89ea28bd940388010606302fa97f4193efa0b0cf/src/publish.coffee#L79) is not enough, especially when Atom has to authenticate with GH.  https://discuss.atom.io/t/cant-publish-package/4890/16?u=joshcheek

### Alternate Designs

I did not consider alternatives as I don't know how to reproduce the original issue as I don't know the cause for a fact.

### Benefits

This will still work as soon as the tag is found, but will longer if it hasn't been found yet. This should be a rare use case, so I don't expect it to add much overhead, but when it goes wrong, it's difficult to figure out what to do (perhaps a --retry) flag, so increasing the time limit will save a lot of frustration in that case (assuming the it just took longer than the timeout).

### Possible Drawbacks

1. While I suspect that the timeout duration is the cause of the issue linked above,
   I do not know this for a fact (I haven't reproduced it as I don't know the cause)
2. No tests exercise this code. On `./bin/npm test`, all tests pass with this change
   but no tests fail when I add a throw statement on the line above. Seems benign enough,
   but maybe look it over more seriously for that reason

### Applicable Issues

* A list of users with this issue + desc of what I noticed when I hit it https://discuss.atom.io/t/cant-publish-package/4890/16?u=joshcheek
* This person hit it, it "resolved itself", so probably a longer timeout would have prevented them from having the issue, saving all the work they put into trying to fix it, and the maintainers getting an issue opened when it wasn't really broken https://github.com/atom/apm/issues/670
* +1 on [this](https://github.com/atom/apm/issues/558) issue, if this happened, it would probably not be so painful on a failure here (meaning I'd have immediately repushed, accomplishing the same thing as a timeout. That said, making it atomic is nontrivial as it's already pushed the tags to github.
* +1 to [this](https://github.com/atom/apm/issues/459) issue, as well, I didn't see any way to tell it to try again until I found and read through [this](https://discuss.atom.io/t/cant-publish-package/4890/16?u=joshcheek) whole thread, and then when I did finally understand to use the `--tag` option, I used it wrong (I did `--tag 2.2.2` rather than `--tag v2.2.2`), it seems there are dozens of people who all want to do this but have to go to the internet to find it, and even then it's error prone. Increasing the timeout will hopefully reduce the number of people, but still, why not add a `--retry` flag that maps to `--tag {currentVersion}` 
* Probably would fix this issue: https://github.com/atom/apm/issues/422
* Note that there are two more pages of issues about publishing that I haven't yet looked through (it's a bit tedious, and I feel like the point is already made).

### DIFF
--- a/src/publish.coffee
+++ b/src/publish.coffee
@@ -76,7 +76,7 @@ class Publish extends Command
   # Check for the tag being available from the GitHub API before notifying
   # atom.io about the new version.
   #
-  # The tag is checked for 5 times at 1 second intervals.
+  # The tag is checked for 10 times at 1 second intervals.
   #
   # pack - The package metadata.
   # tag - The tag that was pushed.
@@ -84,7 +84,7 @@ class Publish extends Command
   #            or the maximum numbers of requests for the tag have been made.
   #            No arguments are passed to the callback when it is invoked.
   waitForTagToBeAvailable: (pack, tag, callback) ->
-    retryCount = 5
+    retryCount = 10
     interval = 1000
     requestSettings =
       url: "https://api.github.com/repos/#{Packages.getRepository(pack)}/tags"


50Wliu:
@JoshCheek There was a transient issue with GitHub 2-3 days ago where newly-pushed tags were showing up on the website after a 2-minute delay.  Other than the aforementioned timespan, I've never had issues publishing packages, and even with this change, it still wouldn't have changed anything.

/cc @lee-dohm

lee-dohm:
There are a lot of known issues with publishing, it is true. This change won't solve any of them, unfortunately, as @50Wliu pointed out. I'd rather that we spend our time fixing the actual problems, such as your suggested `--retry` flag would do. I would be happy to review a PR for that change.

Thanks very much for the contribution and your interest in Atom!

PR #693, opened 2017-03-02
Stop using asar archives

as-cii:
As illustrated in https://github.com/atom/atom/pull/13916, we don't need to produce an asar bundle in Atom anymore. This pull request changes `apm` to use the non-asar locations when determining the resource paths.

### DIFF
--- a/src/apm.coffee
+++ b/src/apm.coffee
@@ -26,28 +26,24 @@ module.exports =
 
     apmFolder = path.resolve(__dirname, '..')
     appFolder = path.dirname(apmFolder)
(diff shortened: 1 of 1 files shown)


jsolisu:
Please check. This fails with Atom on master. Settings-View package cannot get list of core packages.

as-cii:
> Please check. This fails with Atom on master. Settings-View package cannot get list of core packages.

As mentioned above, this pull request is required for startup snapshots in Atom because we will stop packaging core modules and resources into an .asar bundle. Everything should work again as expected as soon as https://github.com/atom/atom/issues/13916 gets merged.

PR #716, opened 2017-04-28
npm version bump

smashwilson:
Updating the bundled version of npm to avoid npm/npm#13256, which was breaking the installation of packages with native dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "keytar": "^4.0",
     "mv": "2.0.0",
     "ncp": "~0.5.1",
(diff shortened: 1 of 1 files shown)


smashwilson:
@BinaryMuse Any reason not to cut v1.19 with this fix and roll it in with atom/atom#14269? This seems fairly low-risk.

BinaryMuse:
@smashwilson Hm, the version bump would include another change to apm so let's just do it in a separate PR so we can keep the concerns detangled.

PR #717, opened 2017-04-29
add `apm link --name` option

UziTech:
### Description of the Change

Add a `--name` option to `apm link`

```
apm link [<package_path>] [--name <package_name>]
```

### Alternate Designs

We could also add the `-n` short option

### Benefits

- Set the name of a linked package easily
- Being able to have two versions of the same package (Most likely one of them disabled based on context)
- Overriding a core package with a local package

Right now the way to do this is to change the `name` field in `package.json`, then link the package, then change the `name` field back.

### Possible Drawbacks

none?

### Applicable Issues

none

### DIFF
--- a/spec/link-spec.coffee
+++ b/spec/link-spec.coffee
@@ -155,3 +155,30 @@ describe 'apm link/unlink', ->
 
       runs ->
         expect(fs.existsSync(path.join(atomHome, 'packages', path.basename(numericPackageName)))).toBeFalsy()
+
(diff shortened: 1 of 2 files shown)


UziTech:
@50Wliu can I get this reviewed?

UziTech:
Any feedback on this?

UziTech:
🙃

smashwilson:
👍 Sure, this would be handy.

PR #623, opened 2016-10-18
add --enabled and --disabled flags to `apm list`

bronson:
As requested in https://github.com/atom/apm/pull/400#issuecomment-253913575

This is based on top of #617 so that PR should be merged first.

PR #729, opened 2017-07-18
Textmate conversion error reporting & apm list --enabled

FrederickGeek8:
#### Change 1: Textmate Bundle Conversion Error Reporting
My change is in response to atom/first-mate#51 – an issue which requested that Atom Package Manager give more information, such as file names and line numbers, in regards to errors when converting TextMate Bundle files to an Atom digestible format. 

I decided to apply the fix in apm instead of first-mate since the information required to locate the source of the error is not made available to first-mate. Additionally, it appears to be near impossible (or at least efficient) to ascertain an exact location of an error, so I elected to output the malformed piece of text instead, which should hopefully give enough information to the user to debug their program.

#### Change 2: New `apm list` command
Due to a mistake on my part, instead of opening a new pull request for this change, I accidentally merged it into this pull request. The second change is one that fulfills #732 – an issue that requests a feature that allowed for the command `apm list --installed --enabled` to list enabled packages in Atom.

### Alternate Designs

My first thought was to feed the _scope_ variable (that had a chance to be malformed, causing the crash), back into a regex on the original file, allowing the error position to be ascertained. However, given how the Package Converter is structured, it would be too computationally intensive to perform such an action.

No alternate design was considered for the second change. I do not see any method as simple and concise as the approach I took

### Benefits

Better error reporting during TextMate Bundle conversion.

### Possible Drawbacks

I see no possible drawbacks to this change.

### Applicable Issues

atom/first-mate#51

### DIFF
--- a/spec/list-spec.coffee
+++ b/spec/list-spec.coffee
@@ -110,9 +110,9 @@ describe 'apm list', ->
 
     listPackages ['--enabled'], ->
       expect(console.log.argsForCall[1][0]).toContain 'test2-module@1.0.0'
-      expect(console.log.argsForCall[3][0]).toContain 'dev-package@1.0.0'
-      expect(console.log.argsForCall[6][0]).toContain 'user-package@1.0.0'
-      expect(console.log.argsForCall[9][0]).toContain 'git-package@1.0.0'
+      expect(console.log.argsForCall[4][0]).toContain 'dev-package@1.0.0'
+      expect(console.log.argsForCall[7][0]).toContain 'user-package@1.0.0'
+      expect(console.log.argsForCall[10][0]).toContain 'git-package@1.0.0'
 
   it 'lists packages in json format when --json is passed', ->
     listPackages ['--json'], ->
--- a/spec/list-spec.coffee
+++ b/spec/list-spec.coffee
@@ -88,7 +88,19 @@ describe 'apm list', ->
     listPackages [], ->
       expect(console.log.argsForCall[1][0]).toContain 'test-module@1.0.0 (disabled)'
 
-  it 'displays only enabled packages --enabled is called', ->
+  it 'displays only enabled packages when --enabled is called', ->
+    atomPackages =
+      'test-module':
+        metadata:
+          name: 'test-module'
+          version: '1.0.0'
+      'test2-module':
+        metadata:
+          name: 'test2-module'
+          version: '1.0.0'
+
+    fs.writeFileSync(path.join(resourcePath, 'package.json'), JSON.stringify(_atomPackages: atomPackages))
+
     packagesPath = path.join(atomHome, 'packages')
     fs.makeTreeSync(packagesPath)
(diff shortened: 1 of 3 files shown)


src/list.coffee
```diff
@@ -81,11 +83,11 @@ class List extends Command
       manifest ?= {}
       manifest.name = child
       if options.argv.themes
-        packages.push(manifest) if manifest.theme
+        packages.push(manifest) if manifest.theme and not (options.argv.enabled and @isPackageDisabled(child))
```
50Wliu: I think these inline ifs are getting a bit hard to understand at a glance.  Would you mind converting them into regular ifs? Eg
```coffee
if mainfest.theme and not (options.argv.enabled and @isPackageDisabled(manifest.name))
```
Also, `manifest.name` is clearer in its intent than just `child`.

src/package-converter.coffee
```diff
@@ -162,7 +162,12 @@ class PackageConverter
         extension = path.extname(child)
         name = path.basename(child, extension)
 
-      selector = new ScopeSelector(scope).toCssSelector() if scope
+      try
+        selector = new ScopeSelector(scope).toCssSelector() if scope
+      catch e
+        e.fileName = path.join(sourceSnippets, child)
+        e.message = "In file " + e.fileName + " at " + JSON.stringify(scope) + ": " + e.message
```
50Wliu: What do you think about using interpolation here instead?

src/package-converter.coffee
```diff
@@ -162,7 +162,12 @@ class PackageConverter
         extension = path.extname(child)
         name = path.basename(child, extension)
 
-      selector = new ScopeSelector(scope).toCssSelector() if scope
+      try
+        selector = new ScopeSelector(scope).toCssSelector() if scope
+      catch e
+        e.fileName = path.join(sourceSnippets, child)
```
50Wliu: The `fileName` property seems to be non-standard and only works in Firefox.  I'm not sure we need this considering the filename is printed anyway in the message.

FrederickGeek8:
@50Wliu I changed what you requested and added some Jasmine tests.

spec/list-spec.coffee
```diff
@@ -88,6 +88,17 @@ describe 'apm list', ->
     listPackages [], ->
       expect(console.log.argsForCall[1][0]).toContain 'test-module@1.0.0 (disabled)'
 
+  it 'displays only enabled packages --enabled is called', ->
```
50Wliu: Think you forgot a `when` here :)

Also, can you also copy over a different (enabled) package just to confirm that the output isn't always `(empty)`? :grinning:

FrederickGeek8: Sorry about that... Should be fixed now as well as a layer of verification that the package field is not always empty.

FrederickGeek8:
@50Wliu just a heads up, I made the requested changes

50Wliu:
Thanks!

bronson:
Hi, --enabled was already available in https://github.com/atom/apm/pull/623

Curious why that one was never reviewed or merged.

PR #731, opened 2017-07-21
Always fall back to ATOM_ACCESS_TOKEN when there is no token

zertosh:
### Description of the Change

This fixes two issues related to `keytar`'s recent switch to a Promise-based API (see atom/node-keytar#63). 

### Benefits

* If `keytar.findPassword`'s promise was rejected, there was an uncaught promise rejection exception, instead of falling back to `process.env.ATOM_ACCESS_TOKEN`. That's fixed.
* If `keytar` failed to load, the stub for `findPassword` would return `undefined` instead of a promise. That's fixed.
* The `keytar` stub module had a `replacePassword` method, when it should in fact be `setPassword`. That's fixed.

### Possible Drawbacks

None.

### Applicable Issues

Nuclide publishes from CircleCI, which is where this issue was found. See `./scripts/oss-publish.sh` in https://circleci.com/gh/facebook/nuclide/1537

```
(node:18872) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Address element '/dev/null' does not contain a colon (:)
```

### DIFF
--- a/src/auth.coffee
+++ b/src/auth.coffee
@@ -4,8 +4,8 @@ catch error
   # Gracefully handle keytar failing to load due to missing library on Linux
   if process.platform is 'linux'
     keytar =
-      findPassword: ->
-      replacePassword: ->
+      findPassword: -> Promise.reject()
+      setPassword: -> Promise.reject()
   else
     throw error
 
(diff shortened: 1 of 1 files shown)


smashwilson:
:+1:

PR #738, opened 2017-08-15
Revert "Stop using asar archives"

as-cii:
Reverts atom/apm#693, since we turned .asar archives back on as part of https://github.com/atom/atom/pull/14682.

@nathansobo: I believe this will fix the `Could not determine Electron version` error we were observing when running `apm install`.

/cc: @iolsen

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "atom-package-manager",
   "description": "Atom package manager",
-  "version": "1.18.3",
+  "version": "1.18.4-0",
   "license": "MIT",
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


virtuoushub:
Not sure if this is related, but I ran into https://github.com/atom/atom/issues/6604 when testing for https://github.com/atom/apm/pull/745.

If I get some time, I want to rebase my PR ontop of 05748001705c8b3a3febd610d5df1fe826c89949 (merge of https://github.com/atom/apm/pull/693), to see if I have to use the [`getResourcePath`](https://github.com/atom/apm/blob/a6285bc4e90e442906bc86542f885f79e2dea15e/src/apm.coffee#L23-L51) workaround I mention.

Ben3eeE:
I think the asar was added back in Atom 1.21 so if you are testing the master version of apm using Atom 1.20 you would run into problems.

virtuoushub:
@Ben3eeE thanks for the insight. I was testing with Atom 1.20

Issue #600, opened 2016-08-09
Duplicate package after renaming


idleberg:
I have recently renamed my Inno Setup package from `language-inno` to `language-innosetup`. The renaming itself, as well as the redirection of the package page worked fine. However, after updating my package locally, I now have two instances of the package installed, one under the old name `language-inno` and another one under the new name `language-innosetup`. Since the files of the old package remain unchanged, the package will continue to request updates.

On my personal computer, I simply deleted the old package, but since I received a [first issue](https://github.com/idleberg/atom-language-innosetup/issues/1#issuecomment-238435874), I wanted to post it here (where it belongs!)
## Steps to Reproduce
- Clone repository, checkout to a tag before package was renamed:

``` bash
cd ~/.atom/packages
git clone https://github.com/idleberg/atom-language-innosetup language-inno
cd language-inno
git checkout tags/v2.1.6
```
- Launch Atom
- Update packages
- Browse to packages directory

**Expected behavior**: [What you expected to happen]

`language-inno` should have been removed

**Actual behavior**: [What actually happened]

Package directory contains two versions of the package: `language-inno` (old) and `language-innosetup` new
## Versions

apm  1.10.0
npm  2.13.3
node 0.10.40
python 2.7.10
git 2.7.4

Atom    : 1.9.6
Electron: 0.37.8
Chrome  : 49.0.2623.75
Node    : 5.10.0

50Wliu:
Should be fixed by #747.

PR #747, opened 2017-09-22
Rename uninstall old module

UziTech:
### Description of the Change

Remove old package if an installed package is renamed

### Benefits

Won't have to manually uninstall the old package if a package is renamed

### Possible Drawbacks

Could break someones [workflow](https://xkcd.com/1172/)

### Applicable Issues

#600

### DIFF
--- a/src/install.coffee
+++ b/src/install.coffee
@@ -356,7 +356,7 @@ class Install extends Command
         commands.push (packagePath, next) =>
           @installModule(options, pack, packagePath, next)
         if installGlobally and packageName isnt pack.name
-          commands.push (newPack, next) => # remove package with old name
+          commands.push (newPack, next) => # package was renamed; delete old package folder
             fs.removeSync(path.join(@atomPackagesDirectory, packageName))
             next(null, newPack)
         commands.push ({installPath}, next) ->
--- a/src/install.coffee
+++ b/src/install.coffee
@@ -355,6 +355,10 @@ class Install extends Command
             @installNode (error) -> next(error, packagePath)
         commands.push (packagePath, next) =>
           @installModule(options, pack, packagePath, next)
+        if installGlobally and packageName isnt pack.name
+          commands.push (newPack, next) => # remove package with old name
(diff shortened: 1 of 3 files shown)


src/install.coffee
```diff
@@ -355,6 +355,10 @@ class Install extends Command
             @installNode (error) -> next(error, packagePath)
         commands.push (packagePath, next) =>
           @installModule(options, pack, packagePath, next)
+        if installGlobally and packageName isnt pack.name
+          commands.push (newPack, next) => # remove package with old name
```
50Wliu: I would make this comment a bit clearer; something like "package was renamed; delete old package folder"

50Wliu:
I can't tell for sure, but it looks like Travis might be failing with this change?

UziTech:
It seems to be failing on one of the delete specs which this shouldn't affect. When just running the install specs everything works.

src/install.coffee
```diff
@@ -355,6 +355,10 @@ class Install extends Command
             @installNode (error) -> next(error, packagePath)
         commands.push (packagePath, next) =>
           @installModule(options, pack, packagePath, next)
+        if installGlobally and packageName isnt pack.name
+          commands.push (newPack, next) => # package was renamed; delete old package folder
+            fs.removeSync(path.join(@atomPackagesDirectory, packageName))
```
UziTech: I looked into it further and it looks to be having an issue with this line. More specifically the `fs.removeSync` although I can't seem to figure out why it is throwing an error.

50Wliu: You can probably wrap this in a try/catch and log the error.

UziTech: Tried that. It doesn't display an error. It won't even let jasmine finish testing all the specs.

UziTech: Finally figured it out. The test-module2 fixture was giving the wrong name so it was failing on other tests.

UziTech:
@50Wliu Could we get a new release for this?

damieng:
Cause of issue https://github.com/atom/settings-view/issues/1018

PR #752, opened 2017-11-07
Bring bundled Node.js version in line with Atom's

smashwilson:
apm's bundled Node.js is currently a major version behind Atom's:

```
smashwilson @ hubtop ~/src/atom (aw-watcher *) [1] 
$ apm --version
apm  1.18.10
npm  3.10.10
node 6.9.5 x64
atom 1.24.0-dev
python 2.7.10
git 2.15.0
smashwilson @ hubtop ~/src/atom (aw-watcher *) 
$ atom --version
Atom    : 1.24.0-dev-3f264fa2b
Electron: 1.6.15
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

This is causing problems for me because I'm using async/await in an install script for `@atom/watcher`, which requires [Node.js >= 7.0.0](http://node.green/). But, in general, it feels weird to have two different Node.js versions in apm and Atom proper.


smashwilson:
Obsoleted by #796.

PR #764, opened 2017-12-19
Make rename detection case-insensitive

damieng:
Fixes https://github.com/atom/settings-view/issues/1018


50Wliu:
Should we add a few tests for uppercase package names to make sure we handle those correctly going forward?

PR #773, opened 2018-02-10
Allow installing old versions not listed in the package info

### LINKED ISSUE #733: apm can't install a version that is more than [pagination limit] behind the latest
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Running `apm install autocomplete-python@1.8.11` results in an error even though the version is listed in the database. 

### Steps to Reproduce

1. Open a terminal session
2. Execute `apm install autocomplete-python@1.8.11`

**Expected behavior:** autocomplete-python v1.8.11 to be installed

**Actual behavior:** `Package version: 1.8.11 not found` error returned

### Versions

```
Atom    : 1.20.0-dev-57e023e74
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

```
apm  1.18.3
npm  3.10.10
node 6.9.5 x64
python 2.7.13
git 2.13.3
```

### Additional Information

I suspect this is because the results from the API aren't paginated.

alangpierce:
### Description of the Change

As described in #733, installation currently makes a request to `GET /api/packages/:package_name` and uses the `versions` key to get the tarball URL. This only works with recent versions because the API only returns the 30 most recent versions here. To allow installing older versions, this change makes it so we first check the package information, then make another request to `GET /api/packages/:package_name/versions/:version_name` if necessary.

Since I changed the code to use `version` to mean a version JSON object, I renamed `packageVersion` to `versionName` in a few places to be more clear. I also rely on `version.version` containing the version name, which required updating some test fixtures.

This also adds tests for installing old versions in general and producing the expected error message if a version isn't found.

### Alternate Designs

Another approach would be to modify the `GET /api/packages/:package_name` to either support pagination or return all versions. The current approach is nice because it's just a change to the client and does not have the risk of a performance regression in the common case. It's also simpler than trying to use a paginated endpoint.

Yet another approach would be to derive the tarball URL directly rather than getting it from the API, since it seems to follow a standard format. This would simplify the code, but would be less HATEOAS-friendly and make it harder to change tarball URLs in the future.

### Benefits

This change makes it possible to install old versions of packages. In particular, I'm getting a build failure in the [decaffeinate test suite](https://github.com/decaffeinate/decaffeinate#status) because I'm trying to build Atom from last September.

### Possible Drawbacks

Installing a nonexistent version will now be slightly slower because it will do another API request.

### Applicable Issues

Fixes #733


alangpierce:
See also this Slack conversation: https://atomio.slack.com/archives/C044E54H0/p1517937415000412
cc @lee-dohm

lee-dohm:
Thanks for this @alangpierce! It looks pretty good at first glance and I'll dig into it on Monday :+1:

lee-dohm:
Would you mind fixing the conflicts here?

PR #775, opened 2018-03-05
Improve Python 2 detection

smashwilson:
Rather than strictly relying on Python 2 being available on your ${PATH} with a binary name of `python`, use [@atom/dowsing-rod](https://github.com/atom/dowsing-rod) to locate the best Python binary and trick node-gyp into using it.

See atom/atom#16885 for more information.

### DIFF
--- a/bin/python-interceptor.sh
+++ b/bin/python-interceptor.sh
@@ -34,9 +34,9 @@ case $1 in
       ARGS+=("--format=safemake.py")
     fi
 
-    exec python "${ARGS[@]}"
+    exec "${ATOM_APM_ORIGINAL_PYTHON:-python}" "${ARGS[@]}"
     ;;
   *)
-    exec python "$@"
+    exec "${ATOM_APM_ORIGINAL_PYTHON:-python}" "$@"
     ;;
 esac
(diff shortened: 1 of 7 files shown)


bin/python-interceptor.sh
```diff
@@ -34,9 +34,9 @@ case $1 in
       ARGS+=("--format=safemake.py")
     fi
 
-    exec python "${ARGS[@]}"
+    exec "${ATOM_APM_ORIGINAL_PYTHON:-python}" "${ARGS[@]}"
```
smashwilson: I could swear I've fixed this before...

Sighery:
Seems like `node-gyp` accepts Python3 now. It says so on [their README](https://github.com/nodejs/node-gyp). And judging by the commit history it seems they started working on it somewhere around July of this year. Not sure what's the status of this branch, but seeing as it's open I figured it was a good idea to give a heads up.

DeeDeeG:
This is maybe obsolete now that #887 is merged.

Though the `apm --version` does need some help finding `python2` and `python3`, could be better at finding Python on Windows.

PR #252, opened 2015-01-13
Upgrade npm

kevinsawicki:
Upgrades to the latest 2.1.x version of npm

Closes #245
Closes #251

PR #796, opened 2018-07-11
Massive node and npm version bumps

smashwilson:
Node and npm have advanced quite a bit since the last time that we really spent time on apm. As a result, apm has lagged _significantly_ behind the state-of-the-art.

In this pull request I'm upgrading:

* Node.js to v8.9.3, to match the version that Atom's current Electron version gives us; and
* npm to 6.1.0, to give our package authors access the latest-and-greatest as of now.

The npm bump in particular will give us access to features like:

* Proper `package-lock.json` support
* `audit` and `audit fix`
* :rocket: mucho performance boost

However, this is _also_ super risky. We're yanking npm three full major versions here. We'll have to do some thinking about the best way to roll this out. I'm also tinkering with this "on the side" (meaning, while other things are waiting for feedback) rather than as my primary focus. So if you're watching the repo, don't get your hopes up for a speedy fix and ship 😅.

/cc @daviwil for general awareness


spec/clean-spec.coffee
```diff
@@ -24,24 +24,35 @@ describe 'apm clean', ->
...
-    moduleDirectory = path.join(temp.mkdirSync('apm-test-module-'), 'test-module-with-dependencies')
-    wrench.copyDirSyncRecursive(path.join(__dirname, 'fixtures', 'test-module-with-dependencies'), moduleDirectory)
-    process.chdir(moduleDirectory)
+      moduleDirectory = path.join(temp.mkdirSync('apm-test-module-'), 'test-module-with-dependencies')
+      wrench.copyDirSyncRecursive(path.join(__dirname, 'fixtures', 'test-module-with-dependencies'), moduleDirectory)
+      process.chdir(moduleDirectory)
+      live = true
+    waitsFor -> live
 
   afterEach ->
```
50Wliu:
```coffee
waitsFor (done) ->
  server.close -> done()
```
*should* work.

daviwil:
Woohoo!  Will be following along closely :)

spec/clean-spec.coffee
smashwilson: Oh, good catch.

smashwilson: ... Nope:

```
TypeError: done is not a function
  at Server.<anonymous> (/Users/smashwilson/src/atom/apm/spec/clean-spec.coffee:40:9)
  at Object.onceWrapper (events.js:313:30)
  at emitNone (events.js:106:13)
  at Server.emit (events.js:208:7)
  at emitListeningNT (net.js:1378:10)
  at _combinedTickCallback (internal/process/next_tick.js:135:11)
  at process._tickCallback (internal/process/next_tick.js:180:9)
```

I bet that's implemented in a different Jasmine helper library? I know I've seen a few floating around.

smashwilson:
apm makes assumptions about the internal structure of npm's cache:

https://github.com/atom/apm/blob/478f7be389cdb1bf38c6f284d7631cad04e62e91/src/install.coffee#L271-L290

At some point, npm [rewrote its caching implementation completely](https://docs.npmjs.com/cli/cache#details), and it would be A Bad Idea to depend on its internal structure again:

> There is currently no method exposed through npm to inspect or directly manage the contents of this cache. In order to access it, cacache must be used directly.

For the moment, I've worked around this by shunting the tarball cache to an `.apm/tarballs/` subdirectory. This is probably a bad idea because it means we'll be caching apm packages twice, which would likely cause a bunch of bloat. I'd rather shunt _all_ of the cache handling to npm and, if possible, do the `npm install` call with a tarball address directly rather than doing the tarball download ourselves. I'm not there yet, though - right now I'm just getting the tests clean.

daviwil:
Sounds like a solid approach if it works! :crossed_fingers:

smashwilson:
I have a bad feeling that along the way I'll rediscover the reason why we didn't just do that in the first place. Worth a shot though :wink:

smashwilson:
Doing some blame spelunking, it looks like the manual cache management was added in... #24. And then touched again in #252, during the upgrade to npm _2.1.18_. It's quite possible that npm didn't accept a tarball URL at all yet.

I've yanked it out and am just passing the tarball URL through now; it was pretty easy and all of my tests are still green.

smashwilson:
This is green everywhere now? I just pushed a prerelease tag (`v1.19.0-0`) so we can try it out in Atom proper.

smashwilson:
Hahaha. Fun fact about more recent npm versions: they automatically prune your `node_modules/` of anything not mentioned in your `package.json`. In the case of the Atom build, that means "absolutely everything but the most recent apm-installed package". So every bundled package deletes all of the others.

I'm looking through npm's source to see if there's a way to prevent `npm install` from performing an automatic prune 🔎

smashwilson:
Spoiler: no. You can prevent autopruning with `--no-package-lock`, but that _also_ prevents the use of `package-lock.json` files entirely.

We'll need to do some thinking about what we actually want `apm install` to do with lockfiles in packages with `packageDependencies` specified 🤔

thomasjo:
Looks like we might be able to resolve the auto-pruning of atom packages by augmenting those package's `package.json` file in the `node_modules` directory. By adding 
``` json
"_requiredBy": ["/"] // Alternatively "#USER"
```
we should be able to convince npm that these packages are actually required. I discovered this by investigating the logic behind [is-extraneous.js](https://github.com/npm/npm/blob/4c65cd952bc8627811735bea76b9b110cc4fc80e/lib/install/is-extraneous.js#L15-L26) in the npm source.

smashwilson:
I'm leaning toward just checking in the tarball URLs added to `package.json` at the moment. Finding ways to remove them feels too much like fighting npm's design rather than working with it. Much of npm's core assumes that `package.json` is the source of truth for what the `node_modules/` directory should contain, and the farther we deviate from that, the more difficult our life will be moving forward.

Some of the benefits of letting npm do its thing:

* Atom's packageDependencies and their transitive dependencies will be centrally managed in the atom/atom build. This will let us run `npm audit fix` on the entire atom/atom dependency tree at once, and may lead to space savings once npm can deduplicate across bundled packages.
* Running npm commands like `npm link` in the atom/atom tree will no longer accidentally prune all of the packages. I've been bitten by this one a bunch, it basically makes `npm link` unusable in Atom.
* I'm not 100% certain, but it might open the door to letting us use `npm ci` in Atom's build, which would be a big build performance boost.
* We don't increase our maintenance burden the next time we upgrade npm. We'd need to figure this all out again if npm changes its pruning logic.

The biggest downsides I can think of:

* It's kind of ugly.
* It'd be harder to edit Atom's `package.json` in the web editor to do package bumps.

Is there anything else I'm not thinking of?

daviwil:
Does this only affect the `package.json` of Atom and not community packages?  If it's only for Atom, I'm pretty fine with that.  Could you give an example of what it adds to `package.json`?

smashwilson:
> Could you give an example of what it adds to package.json?

You can see what it adds in [this commit](https://github.com/atom/atom/pull/17675/commits/6dabeffd25cc80345c6051ec3d4082bd460aae35).

> Does this only affect the `package.json` of Atom and not community packages? If it's only for Atom, I'm pretty fine with that.

Strictly speaking, it'll affect the `package.json` of any project that uses the apm-specific `packageDependencies` field. I think community packages _can_ use that but I'm not sure how widely it's used (and registering and consuming services would be a better pattern anyway). Let me see if how many people are using that...

smashwilson:
Here's the list of packages that would be effected:

```
$ find . -type f -name package.json -print0 | xargs -0 -- grep packageDependencies
./greyatom-ide/package.json:  "packageDependencies": {
./gradle-manager/package.json:  "packageDependencies": {
./awesome-tool-bar/package.json:  "packageDependencies": {
./toolbar-basic/package.json:  "packageDependencies": {
./bloginatom/package.json:  "packageDependencies": {
./compiler-invoker/package.json:  "packageDependencies": {
./tool-bar-besser/package.json:  "packageDependencies": {
./tool-bar-shortcuts/package.json:  "packageDependencies": {
./language-emp-view/package.json:  "packageDependencies": {
./emp-debugger/package.json:  "packageDependencies": {
./tool-bar-config/package.json:  "packageDependencies": {
./grunt-manager/package.json:  "packageDependencies": {
./atom-imp-tools/package.json:  "packageDependencies": {
./tool-bar-almighty/package.json:  "packageDependencies": {
./dev-greyatom-ide/package.json:  "packageDependencies": {
./tool-bar-custom/package.json:  "packageDependencies": {
./frontend-tool-bar/package.json:  "packageDependencies": {
./test-dp-1/package.json:  "packageDependencies": {
./tool-bar-atom/package.json:  "packageDependencies": {
./ioton/package.json:  "packageDependencies": {
./linter-manager/package.json:  "packageDependencies": {
./learn-ide-3/package.json:  "packageDependencies": {
./todo-manager/package.json:  "packageDependencies": {
./mastermind/package.json:  "packageDependencies": {
./tool-bar-bgc/package.json:  "packageDependencies": {
./gulp-manager/package.json:  "packageDependencies": {
./toolbar-fusion/package.json:  "packageDependencies": {
./toolbar-iconshortcuts/package.json:  "packageDependencies": {
./tool-bar-hayatan/package.json:  "packageDependencies": {
```

daviwil:
Ahhh, I see.  Not too bad, I guess, just a bit of duplication to reference the tarball URLs explicitly.  I'm hoping that PR https://github.com/atom/atom/pull/17686 can remove the necessity of using most of our bundled core packages from apm anyway.

PR #797, opened 2018-07-13
Add support for repo-local packageDependencies in Atom's package.json

daviwil:
### Description of the Change

This change adds support for npm's `file:` syntax for `packageDependencies` in Atom's `package.json` file.  When a package dependency uses such a string with a relative path, we use a special code path that knows to pass along that path to npm so that the package can be installed from the local folder into `node_modules`.

### Alternate Designs

I don't know of any other way to accomplish this with `apm` currently, but I'd be happy to learn of alternatives!

### Benefits

This allows Atom to easily use core packages which live inside a subfolder of Atom's own repository instead of having to ship core packages to apm and then reference them with version numbers in `packageDependencies`.

### Possible Drawbacks

Currently there's no way to determine that a package from a local path and its "installed" copy under `node_modules` are the same and don't require another install.  This is mainly due to the fact that we no longer change the version in the package's `package.json` file, so it's hard to know for certain whether the package code has changed between builds.

### Applicable Issues

- RFC 003: https://github.com/atom/atom/pull/17674
- PR to Atom that consumes this change: https://github.com/atom/atom/pull/17686

### DIFF
--- a/src/install.coffee
+++ b/src/install.coffee
@@ -307,6 +307,45 @@ class Install extends Command
               @logSuccess() unless options.argv.json
           callback(error, json)
 
+  # Install the package with the given name and local path
+  #
+  # packageName - The name of the package
+  # packagePath - The local path of the package in the form "file:./packages/package-name"
+  # options     - The installation options object.
+  # callback    - The function to invoke when installation completes with an
+  #               error as the first argument.
+  installLocalPackage: (packageName, packagePath, options, callback) ->
+    unless options.argv.json
+      process.stdout.write "Installing #{packageName} from #{packagePath.slice('file:'.length)} "
+
+    @requestPackage packageName, (error, pack) =>
+      if error?
+        @logFailure()
+        callback(error)
+      else
+        # packageVersion ?= @getLatestCompatibleVersion(pack)
+        # unless packageVersion
+        #   @logFailure()
+        #   callback("No available version compatible with the installed Atom version: #{@installedAtomVersion}")
+        #   return
+        commands = []
+        commands.push (next) =>
+          @installModule(options, pack, packagePath, next)
+        commands.push ({installPath}, next) ->
+          if installPath?
+            metadata = JSON.parse(fs.readFileSync(path.join(installPath, 'package.json'), 'utf8'))
+            json = {installPath, metadata}
+            next(null, json)
+          else
+            next(null, {}) # installed locally, no install path data
+
+        async.waterfall commands, (error, json) =>
+          if error?
+            @logFailure()
+          else
+            @logSuccess() unless options.argv.json
+          callback(error, json)
+
   # Install all the package dependencies found in the package.json file.
   #
   # options - The installation options
@@ -318,7 +357,10 @@ class Install extends Command
     for name, version of @getPackageDependencies()
       do (name, version) =>
         commands.push (next) =>
-          @installRegisteredPackage({name, version}, options, next)
+          if version.indexOf('file:.') is 0
+            @installLocalPackage(name, version, options, next)
+          else
+            @installRegisteredPackage({name, version}, options, next)
 
     async.series(commands, callback)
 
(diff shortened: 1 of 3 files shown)


src/install.coffee
```diff
@@ -386,7 +425,10 @@ class Install extends Command
     for name, version of @getPackageDependencies()
       do (name, version) =>
         commands.push (next) =>
-          @installRegisteredPackage({name, version}, options, next)
+          if version.indexOf('file:.') is 0
```
50Wliu: :art: `if version.startsWith('file:.')`

daviwil: Thanks!  I think my brain goes sideways when I start looking at Coffeescript

src/install.coffee
```diff
@@ -318,7 +346,10 @@ class Install extends Command
     for name, version of @getPackageDependencies()
       do (name, version) =>
         commands.push (next) =>
-          @installRegisteredPackage({name, version}, options, next)
+          if version.startsWith('file:.')
```
smashwilson: You could also parse these with Node's [url module](https://nodejs.org/dist/latest-v8.x/docs/api/url.html) and dispatch on `protocol`:

<details>

```
$ node
> const url = require('url')
undefined
> url.parse('file:./foo/bar/baz')
Url {
  protocol: 'file:',
  slashes: null,
  auth: null,
  host: null,
  port: null,
  hostname: null,
  hash: null,
  search: null,
  query: null,
  pathname: './foo/bar/baz',
  path: './foo/bar/baz',
  href: 'file://./foo/bar/baz' }
> url.parse('merge-conflicts')
Url {
  protocol: null,
  slashes: null,
  auth: null,
  host: null,
  port: null,
  hostname: null,
  hash: null,
  search: null,
  query: null,
  pathname: 'merge-conflicts',
  path: 'merge-conflicts',
  href: 'merge-conflicts' }
>
```
</details>

It might be a little more flexible and easier to extend in the future. This should be fine for now, though 👍

daviwil: No problem, I'll add that!

daviwil:
The impact is that apm will try to "install" repo-local packages every time we run script/build, though it likely won't add much extra time to the build since if the repo-local package has already been installed into `node_modules` its dependencies will already have been installed and they won't get installed again.

TL;DR: It could make the build very slightly slower.

jasonrudolph:
> It could make the build very slightly slower.

@daviwil: Thanks for that insight. :bow: If the package has native dependencies that get compiled, will apm re-compile those native dependencies every time `script/build` runs?

daviwil:
It shouldn't happen but I'll try it just to be safe before I merge this!

src/install.coffee
daviwil: On second thought, I think it might be better if we start parsing the version string at an earlier point in the code so that we can attach more metadata to it related to whether it's a version string, url path, etc.  I think I'll defer that for now so that I can get this change merged and get the Atom PR moving forward.

daviwil:
@jasonrudolph I tested two scenarios:

1. **Repo-local package dependency directly containing native code:** `node-gyp` is run every time `apm install` is called regardless of whether the package had been installed previously.  Since we don't seem to have any packages with their own native code (no `binding.gyp` file) this shouldn't be a problem for us.

2. **Repo-local package dependency with a dependency containing native code:**  In this case, the package's dependency containing native bits *does not* get recompiled since the dependency has already been installed.

To put it more simply: the `node_modules` folder of installed repo-local core packages in the developer's local Atom repo gets reused between builds, so native modules they contain will not be recompiled for every `apm install`.  Native modules will only be recompiled on every `apm install` if they are contained directly within the package's codebase (indicated by the presence of a `binding.gyp` file).  This is currently not the case for any core Atom package and *not likely* to ever occur.

Any objections to moving forward with this change?

jasonrudolph:
> Any objections to moving forward with this change?

@daviwil: Thanks for the research and the helpful explanation. 💟

Ship ship ship. 👍 :shipit: :rocket:

daviwil:
Thanks a bunch! :shipit:

PR #802, opened 2018-07-25
Don't unpack array

### LINKED ISSUE #801: Fix 'apm disable'
### Issue

```
$ apm install teletype
$ apm disable teletype
$ apm list
teletype@0.13.3
```

### Expected

1. `apm list` should output `teletype@0.13.3 (disabled)`
2. Package should not work
3. Package should show as disabled in Atom Package Settings (UI)
4. Package should be in `*.core.disabledPackages` Array in `~/.atom/config.cson`.

NOTE: Package UI enable/disable works, as does `apm enable`.

### Versions

[Discussion](https://discuss.atom.io/t/apm-disable-doesnt-work-on-my-machine/57133) with @rsese and @Aerijo: 
**Works**: 1.28.1 & 1.29.0-beta2
**Fails**: 1.28.2 & 1.30 (on master/nightly)

MacOS 10.13.5
apm 1.19.0

Aerijo:
### Description of the Change

Doesn't unpack array of package names when calling [`_.union`](https://underscorejs.org/#union), so it works properly. The original version only worked because of a bug in underscore.js, which was patched some time between version 1.6.0 and 1.8.3 (of _.js). They have now explicitly prevented non-arrays from appearing in the output of a union. I.e., `_.union([1, 2], 3, [4]) === [1, 2, 4]` (and that 3 was the disabled package...)

### Benefits

Actually disable packages

### Possible Drawbacks

I don't know why the spread was there initially. I compiled and tested though, and disabling worked fine.

### Applicable Issues

Fix #801 
Fix atom/atom#17741

### DIFF
--- a/src/disable.coffee
+++ b/src/disable.coffee
@@ -69,7 +69,7 @@ class Disable extends Command
 
       keyPath = '*.core.disabledPackages'
       disabledPackages = _.valueForKeyPath(settings, keyPath) ? []
-      result = _.union(disabledPackages, packageNames...)
+      result = _.union(disabledPackages, packageNames)
       _.setValueForKeyPath(settings, keyPath, result)
 
       try


Aerijo:
... I don't know why appveyor failed. Does it work for the current repo as is?

lee-dohm:
@smashwilson Can you do a quick :eyes: on this?

JBallin:
`disable` is working for me in 1.29.0, so this may no longer be an issue. Haven't tested beta/nightly. I closed my issue #801

smashwilson:
👍 Good catch.

> ... I don't know why appveyor failed. Does it work for the current repo as is?

Great question. It looks like it's failing because of a bunch of lint errors, but it's using the same version of coffeelint that I have locally which lints clean? I'll merge this and look into that separately.

PR #805, opened 2018-08-06
npm audit

smashwilson:
Let's run `npm audit fix --force` and fix a bunch of potential security issues.


package.json
```diff
@@ -51,11 +51,11 @@
   },
   "devDependencies": {
     "coffee-script": "^1.8.0",
-    "express": "~3.2.3",
-    "grunt": "^1.0.1",
-    "grunt-coffeelint": "0.0.16",
+    "express": "^4.16.3",
```
Arcanemagus: This is changing from a `~` range to a `^`. Not necessarily a bad thing, just wanted to bring it up.

smashwilson: Looks like that's a relic from a [https://github.com/atom/apm/commit/a087bf041938fc578a3ca0eca03f41ebb2bd534a](long ago time), when all of the dependencies where ~-ranges.

🤷‍♂️

PR #808, opened 2018-08-08
Prevent unnecessary native module rebuilding

smashwilson:
When the system's host node has the same native module version as the downloaded one, don't rebuild native modules in the postinstall script. This should keep us from building libgit2 and oniguruma twice on every install. ✌️ 

Also: optionally disable automatic npm deduplication during postinstall by setting a `NO_APM_DEDUPE` environment variable. I doubt we'll be able to get away with this in our real builds because of the Windows pathlength thing, but it'll help me narrow down the cause of build failures in atom/atom#17803.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "atom-package-manager",
-  "version": "2.1.0",
+  "version": "2.1.0-0",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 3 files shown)


smashwilson:
Well... `NO_APM_DEDUPE` doesn't seem to actually help, unfortunately. But it shouldn't hurt anything, at least?

PR #813, opened 2018-08-30
Update npm

smashwilson:
npm has published a few minor versions since the last upgrade landed. Here are the changelogs:

* [v6.3.0-next.0](https://github.com/npm/cli/releases/tag/v6.3.0-next.0)
* [v6.3.0](https://github.com/npm/cli/releases/tag/v6.3.0)
* [v6.4.0](https://github.com/npm/cli/releases/tag/v6.4.0)
* [v6.4.1](https://github.com/npm/cli/releases/tag/v6.4.1)

I don't see anything _too_ critical in there, but it'll help us stay on top of upstream bugfixes 🐛 🔨 

##### Remaining work

- [x] Bump ye version
- [ ] Tag a prerelease and send a PR to atom/atom including it
- [ ] Verify that the new version does not break Squirrel

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,15 +4,6 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "JSONStream": {
-      "version": "1.3.3",
-      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.3.tgz",
-      "integrity": "sha512-3Sp6WZZ/lXl+nTDoGpGWHEpTnnC6X5fnkolYZR6nwIfzbxxvA8utPWe1gCt7i0m9uVGsSz2IS8K8mJ7HmlduMg==",
-      "requires": {
-        "jsonparse": "^1.2.0",
-        "through": ">=2.2.7 <3"
-      }
-    },
     "abbrev": {
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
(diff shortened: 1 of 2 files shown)


daviwil:
When I tried 6.3.0 the deduping behavior seemed to be worse and apm's dependencies weren't being flattened as much as with 6.2.0.  Not sure if that was just a fluke, but something worth keeping an eye on

daviwil:
When I tried 6.3.0 the deduping behavior seemed to be worse and apm's dependencies in Atom weren't being flattened as much as with 6.2.0.  Not sure if that was just a fluke, but something worth keeping an eye on

smashwilson:
Oh, right - I thought I remembered you having some problems with that. I should cut a prerelease and try a speculative Atom PR to make sure 6.4.1 works well with Squirrel and such.

Atom's VSTS build caught the deduping problems, right?

smashwilson:
Also: I'm deliberately not including this in the v2.1.1 release I'm bringing to Atom in atom/atom#17953 so that we can ship #812 in the next beta hotfix with minimal risk. This is just a "while I'm in here" sort of change, so no big deal if we need to just punt it for later.

daviwil:
VSTS probably won't give an error unless the paths become much longer.  I was checking the path lengths by installing a build on Windows and using PowerShell to get the lengths of the longest 10 paths.  Once you get a PR going on atom/atom I can check the Windows artifact later and see if we're good.

No worries!  Thanks for keeping this up to date :zap:

smashwilson:
Ahh, got it - I'll ping you on the PR once it's up.

> [..] using PowerShell to get the lengths of the longest 10 paths [..]

Think it would be worth committing this somewhere as a script? I could test on my Windows box downstairs too.

daviwil:
Yep, that'd also save me from having to re-remember it every time ;)

daviwil:
Now that I think about it, might be good to have a path length test in the Windows CI script so that we get some indication of how it changes over time and maybe fail the build if it exceeds a predefined goal length we've set.  I'll put together something for that later!

smashwilson:
image

smashwilson:
Just make sure we don't spend more effort on this than it would take to... fix Squirrel to use Windows long path names. 😆

daviwil:
I think fixing Squirrel is probably going to be more of a pain, but I haven't tried it yet :)

DeeDeeG:
Just popping in to sat this is obsoleted as of #887.

PR #814, opened 2018-08-30
Support "apm install --package-lock-only"

smashwilson:
`npm install` supports a `--package-lock-only` option. If provided, `node_modules/` is left untouched, and npm _only_ creates or updates the `package-lock.json` file to reflect the current state of `package.json`. This is useful for making quick edits to dependency ranges without having to do a full install of a large project (like Atom).

##### Remaining work

- [x] Accept the `--package-lock-only` flag.
- [x] Document it in the command-line usage.
- [x] Pass `--package-lock-only` to the `npm install` command forked for non-packageDependencies.
- [x] Ensure that packageDependencies are handled correctly.

### DIFF
--- a/spec/docs-spec.coffee
+++ b/spec/docs-spec.coffee
@@ -16,13 +16,18 @@ describe 'apm docs', ->
       response.sendFile path.join(__dirname, 'fixtures', 'wrap-guide.json')
     app.get '/install', (request, response) ->
       response.sendFile path.join(__dirname, 'fixtures', 'install.json')
-    server =  http.createServer(app)
-    server.listen(3000)
+    server = http.createServer(app)
 
-    process.env.ATOM_PACKAGES_URL = "http://localhost:3000"
+    live = false
+    server.listen 3000, '127.0.0.1', ->
+      process.env.ATOM_PACKAGES_URL = "http://localhost:3000"
+      live = true
+    waitsFor -> live
 
   afterEach ->
-    server.close()
+    done = false
+    server.close -> done = true
+    waitsFor -> done
 
   it 'logs an error if the package has no URL', ->
     callback = jasmine.createSpy('callback')
--- a/spec/featured-spec.coffee
+++ b/spec/featured-spec.coffee
@@ -15,14 +15,18 @@ describe 'apm featured', ->
       response.sendFile path.join(__dirname, 'fixtures', 'packages.json')
     app.get '/themes/featured', (request, response) ->
       response.sendFile path.join(__dirname, 'fixtures', 'themes.json')
+    server = http.createServer(app)
 
-    server =  http.createServer(app)
-    server.listen(3000)
-
-    process.env.ATOM_API_URL = "http://localhost:3000"
+    live = false
+    server.listen 3000, '127.0.0.1', ->
+      process.env.ATOM_API_URL = "http://localhost:3000"
+      live = true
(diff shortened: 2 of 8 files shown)


smashwilson:
Trying this out in atom/atom, it looks like npm is normalizing `file:` URIs slightly differently:

```diff
diff --git a/package.json b/package.json
index cce9107ba..5916b793b 100644
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "@atom/nsfw": "^1.0.18",
     "@atom/source-map-support": "^0.3.4",
     "@atom/watcher": "1.0.8",
-    "about": "file:packages/about",
+    "about": "file:./packages/about",
     "archive-view": "https://www.atom.io/api/packages/archive-view/versions/0.65.1/tarball",
     "async": "0.2.6",
     "atom-dark-syntax": "https://www.atom.io/api/packages/atom-dark-syntax/versions/0.29.1/tarball",
@@ -130,7 +130,7 @@
     "one-dark-syntax": "https://www.atom.io/api/packages/one-dark-syntax/versions/1.8.4/tarball",
     "one-dark-ui": "https://www.atom.io/api/packages/one-dark-ui/versions/1.12.5/tarball",
     "one-light-syntax": "https://www.atom.io/api/packages/one-light-syntax/versions/1.8.4/tarball",
-    "one-light-ui": "file:packages/one-light-ui",
+    "one-light-ui": "file:./packages/one-light-ui",
     "oniguruma": "6.2.1",
     "open-on-github": "https://www.atom.io/api/packages/open-on-github/versions/1.3.1/tarball",
     "package-generator": "https://www.atom.io/api/packages/package-generator/versions/1.3.0/tarball",
```

smashwilson:
Hacky as hell but it works:

```
smashwilson @ hubtop ~/src/atom/atom (aw/apm-up=)
$ ~/src/atom/apm/bin/apm install --package-lock-only
Installing modules ✓
smashwilson @ hubtop ~/src/atom/atom (aw/apm-up=)
$ git diff
smashwilson @ hubtop ~/src/atom/atom (aw/apm-up=)
$
```

daviwil:
So what if we instead changed `apm` to look for `file:` entries without the `.`?  When I originally wrote that logic in `apm` I didn't realize that newer `npm` didn't need the `.` anymore (I think I got that from the npm docs).  This would remove our need for the extra normalization logic.

The only change in our code would be [looking for `file:` instead of `file:.`](https://github.com/atom/apm/blob/master/src/install.coffee#L349) (while also excluding `file:/` if the path is a fully-qualified file URI).  What do you think?

smashwilson:
👍 I've replace the `.startsWith()` test with a regexp that'll permit `file:aaa/bbb` and `file:./bbb/ccc` but not `file:/absolute/path`.

smashwilson:
Hmm looks like we have some flakes lurking in here. Starting and stopping the express server we use to mock npm.org and atom.io for specs is asynchronous, but not all of the tests treat it as such, which means that the server can randomly fail to start because port 3000 isn't free any more. Maybe I'll tidy that up later if I get the chance 🤔

smashwilson:
Hmm this branch is still failing (consistently, for me) with an EADDRINUSE. I wonder if that's the external symptom from a different failure... ?

Our Jasmine setup makes it really hard to tell which test is actually failing 🤔

PR #831, opened 2019-04-02
Update Node to match Electron 3.1

50Wliu:
### Requirements

### Description of the Change

We include a BUNDLED_NODE_VERSION that matches the Node version of the Electron version Atom is running. With atom/atom#18916, Atom will start to use Node 10.2.0.

### Alternate Designs

Get rid of this file.

### Benefits

apm will stay consistent with Atom.

### Possible Drawbacks

Bugs introduced as a result of the upgrade.

### Verification Process

None yet.

### Applicable Issues

None.


rafeca:
Thanks for the PR @50Wliu !

I'm not super familiar with how `apm` works: can we merge this PR and publish a new version of `apm` that uses node v10 before upgrading to electron v3 or do we need to do both things at the same time?

50Wliu:
> can we merge this PR and publish a new version of `apm` that uses node v10 before upgrading to electron v3

Yeah, this should be fine as long as it's a minor version bump. If it turns out we need to do some bugfixes before the Electron 3 PR lands we can always do patch releases for the previous minor version.

rafeca:
Awesome! Then we'll use the new `apm` minor version from the `electron-v3` branch right?

50Wliu:
Yep :)

Aerijo:
My understanding is that the Node version is not tied to Atom compatibility, as we already pass parameters to build native modules against the right Electron version. It just gets put on the PATH, so is what's used when an npm script runs Node.

If I'm understanding this right, would it make sense to use the current [LTS version](https://nodejs.org/en/), which is 10.16.0?

PR #832, opened 2019-04-02
Remove explicit dependency on node-gyp

50Wliu:
### Requirements

### Description of the Change

This dependency was added in #730 as a way to always know how to access the node-gyp binary. However, npm has been bumped several times since then without touching node-gyp, leaving it quite outdated. I believe it should be fine to remove this dependency and simply rely on the version npm ships with.

### Alternate Designs

Update it to the version npm ships with.

### Benefits

Less packages to maintain. This should also unblock using Visual Studio 2017 build tools.

### Possible Drawbacks

npm's directory flattening scheme may change, which will require us to update the node-gyp search path. I think if it does change, we'll notice it fairly quickly and be able to change the path without too much hassle.

### Verification Process

None yet.

### Applicable Issues

None.


lee-dohm:
Can you publish a pre-release build of Atom using this change so that we can make sure that the entire build process works?

vinkla:
@50Wliu will this PR maybe solve https://github.com/atom/atom/issues/18540?

50Wliu:
Perhaps? I don't really know; apm's internals are quite complicated and I'm still figuring out how things work and why they were designed this way.

50Wliu:
Some of these changes weren't the cleanest, specifically all the node-gyp path updating. I'm wondering if we should just try updating our version of node-gyp rather than depending on npm's.

rafeca:
Thanks for doing this! Since the PR on `atom/atom` using the prerelease version passes, I'm confident about merging it.

@50Wliu can you take care of publishing a stable release afterwards and updating the `atom/atom` PR  to point to the stable release?

PR #836, opened 2019-04-23
Remove wrench, fs-plus in favor of fs-extra

50Wliu:
### Requirements

### Description of the Change

[`wrench`](https://npmjs.com/package/wrench) has been deprecated for a few years in favor of [`fs-extra`](https://npmjs.com/package/fs-extra). However, since both `fs-plus` and `fs-extra` happen to extend `fs`, it would be illogical to use both of them in the same codebase. Therefore, this PR removes both `wrench` and `fs-plus` and uses `fs-extra` instead.

### Alternate Designs

None.

### Benefits

1. Using a better-maintained fs alternative
2. No longer depending on a deprecated package that isn't being updated

### Possible Drawbacks

There were two troublesome functions with this swap:
1. `fs.is(Directory|File)Sync`

`fs-extra` doesn't have any way to easily determine if a path is a directory or file without the possibility of throwing exceptions. I've tried to emulate that as best as I could using `try`s everywhere, but it's not that good-looking and the behavior might be subtly different in a few places (we may now throw where we didn't before).

2. `fs.writeFileSync`

`fs-plus` creates any necessary directories first if the path doesn't exist; `fs-extra` doesn't. I tried to add `fs.mkdirp` calls wherever necessary but I may have missed a few.

### Verification Process

For now, just the specs. This will clearly need some extra real-world stress testing as I _do not_ trust these specs.

### Applicable Issues

N/A


50Wliu:
Checked over all the `writeFileSync`s and can confirm that they'll always be writing in an existing folder :+1:

DeeDeeG:
Is this PR still active? I would be interested in taking a shot at updating this against `master`.

DeeDeeG:
Here is an updated version of this PR (with `master` branch merged in): https://github.com/DeeDeeG/apm/commits/wl-fs-extra-merge-master

CI passes: https://travis-ci.com/github/DeeDeeG/apm/builds/179155533

Please consider reviewing. IMO it would be great to start updating `apm`'s dependencies, and I'd be excited if this got merged! Thanks for considering.

50Wliu:
> Is this PR still active?

Uhh...not really, sorry 😬. If you have a branch that is updated, I'd encourage you to open it as a PR!
I'll close this to reduce confusion.

DeeDeeG:
@50Wliu thanks for the reply! I opened #894 with the updated branch.

PR #839, opened 2019-05-03
Run tests on Appveyor

50Wliu:
Let's get this working.

Functional changes:
* Include the mingw directory in the PATH (this directory contains many of the actual git functions, such as `git-pull`, `git-push`, and `git-upload-pack`).
* Search `env.ProgramW6432` for Git before `env.ProgramFiles`, as `ProgramFiles` is _not_ guaranteed to always point to the x64 Program Files directory.

_Technically_, we should get the mingw path from `git --exec-path` (and this is what the GitHub package does), but that would mean making `git.addGitToEnv(env)` async, which would have been a much larger PR.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -3,7 +3,7 @@
 .DS_Store
 npm-debug.log
 /spec
-bin/node
+/bin/node*
 .travis.yml
 .pairs
 appveyor.yml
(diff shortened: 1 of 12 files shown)


50Wliu:
My initial plan was to try to get this working in order to debug failures in atom/atom#19189, but annoyingly specifying x86 as a platform still seems to use an x64 image...

sadick254:
@50Wliu We have since moved from running tests on other platforms and we are fully embracing Github Actions. For that reason, I don't think this is relevant anymore.

PR #843, opened 2019-05-07
What arch are we?

50Wliu:
**🚨 DO NOT MERGE 🚨**

Trying to debug why `process.arch` doesn't seem to be set to the correct value on x86 Azure Pipelines builds.


50Wliu:
✨ _\~Solved\~_ :sparkles:

Actual PR incoming.

PR #853, opened 2019-06-29
Better unlinking behaviour

### LINKED ISSUE #399: apm unlink reports success when package is missing
```
~$ apm unlink adlknfaldskhalksd
Unlinking /Users/bronson/.atom/packages/adlknfaldskhalksd ✓
```

The checkmark indicates success, but I didn't have a package named adlknfaldskhalksd installed.

I would expect it to complain with something like, "package adlknfaldskhalksd wasn't linked".

Aerijo:
### Requirements

### Description of the Change

Makes a mistake in unlinking much easier to detect. Previously, if the passed file was not a valid symlink it would fail silently (and log it as a sucess). With this change, we rely more on the fact that `unlinkSync` throws an error if it doesn't work. Note this is not a spec / behaviour change; the method was able to throw an error already.

### Alternate Designs

Considered catching non-existence in an else block for `if fs.isSymbolicLinkSync(pathToUnlink)`, but decided it was cleaner and more consistent to rely on the `fs.unlinkSync` errors. Note it still only deletes symlinks; attempting on actual directories will throw an error. I'll try and add specs.

### Benefits

Easily see if the operation worked.

### Possible Drawbacks

None. This does not change the possible behaviours of the function. 

The `unlinkAll` method already checks if each item is a symlink before calling `unlinkPath` on it.

### Verification Process

Tried it with valid and invalid (nonexistent, or an actual directory) symlinks on macOS 10.14.4

### Applicable Issues

closes #399

### DIFF
--- a/spec/link-spec.coffee
+++ b/spec/link-spec.coffee
@@ -195,3 +195,15 @@ describe 'apm link/unlink', ->
       runs ->
         expect(console.error.mostRecentCall.args[0].length).toBeGreaterThan 0
         expect(callback.mostRecentCall.args[0]).not.toBeUndefined()
+
+  describe "when unlinking a path that does not exist", ->
+    it "logs an error and exits", ->
+      callback = jasmine.createSpy('callback')
+      apm.run(['unlink', 'a-path-that-does-not-exist'], callback)
+
+      waitsFor 'waiting for command to complete', ->
+        callback.callCount > 0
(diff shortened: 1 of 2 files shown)


as-cii:
Thanks, @Aerijo! I added a couple of tests to ensure this behavior does not regress in the future.

PR #854, opened 2019-06-29
Resolve `.` as folder name when uninstalling

### LINKED ISSUE #460: `apm rm .` (or uninstall) will uninstall all packages
I've accidentally done this a couple times when trying to remove linked packages for the official versions. Luckily, I have a list of all of my installed packages saved and a script to install all of them (for new computers).

Aerijo:
### Requirements

### Description of the Change

Resolves `.` to the name of the current directory when running `apm uninstall .`. It then continues as if the resolved value was the one passed in the first place.

### Alternate Designs

Exactly what to resolve `.` to is tricky. IMO, resolving to the exact current directory is reasonable, but then it doesn't respect the `--dev` flags and such. It also turns the command into a pseudo `rm -rf` instruction, which could be dangerous (it already acts like this on the `packages` directory, and is very annoying). As this PR is currently, any mistakes are isolated to the `packages` and `dev/packages` directories.

That alternative may work better if it first checks that the current folder is the root of an Atom package. But I still think it's best to let a user manage their directories themselves if it's not inside `.atom`.
 

### Benefits

Will no longer delete the entire `~/.atom/packages` folder. This happened because `path.resolve` was resolving the packages path with the package `.`, which is still the packages path, and that folder was deleted.

### Possible Drawbacks

Not necessarily removing the exact folder this was called from (as describe in Alternatives) may be surprising. I made my case there, but could try adding proper relative path support (with package check guard & making sure the resolved path is in the packages directory).

### Verification Process

Ran it on macOS 10.14.4, and added a test.

### Applicable Issues

closes #460

### DIFF
--- a/spec/uninstall-spec.coffee
+++ b/spec/uninstall-spec.coffee
@@ -3,6 +3,18 @@ fs = require 'fs-plus'
 temp = require 'temp'
 apm = require '../lib/apm-cli'
 
+createPackage = (packageName, includeDev=false) ->
+  atomHome = temp.mkdirSync('apm-home-dir-')
+  packagePath = path.join(atomHome, 'packages', packageName)
+  fs.makeTreeSync(path.join(packagePath, 'lib'))
+  fs.writeFileSync(path.join(packagePath, 'package.json'), "{}")
+  if includeDev
+    devPackagePath = path.join(atomHome, 'dev', 'packages', packageName)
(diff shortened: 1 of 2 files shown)


nathansobo:
:zap: This is a nice improvement, and thanks for leaving the tests better than you found them as well.

PR #856, opened 2019-07-08
Better narrow terminal width support

Aerijo:
### Requirements

### Description of the Change

Changes the wrapping value to account for the terminal width; it will now pick the minimum of 100 (original hardcoded value) or the terminal width, if yargs can detect it (`yargs.terminalWidth()` may return `null`, so we default to 100 in this case too).

### Alternate Designs

na

### Benefits

Cleaner display on narrow terminal windows

### Possible Drawbacks

None

### Verification Process

Manually, with the output of `apm --help`

### Applicable Issues

na

### DIFF
--- a/src/apm-cli.coffee
+++ b/src/apm-cli.coffee
@@ -63,7 +63,7 @@ for commandClass in commandClasses
     commands[name] = commandClass
 
 parseOptions = (args=[]) ->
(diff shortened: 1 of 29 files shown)


as-cii:
Thanks, @Aerijo! ⚡️

Issue #818, opened 2018-09-16
apm config set doesn't work


heavyk:
I tried to set python to python2.7 using apm set: `apm config set python python2.7` and it wouldn't work. checked the apmrc file and it was set too..

found this: https://github.com/atom/apm/issues/179#issuecomment-399618944

and followed the directions. uncommenting that line allows apm config set/get to work properly again

PR #703, opened 2017-03-17
Respect a user's PYTHON or npm_config_python settings

smashwilson:
### Requirements

* Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* All new code requires tests to ensure against regressions

### Description of the Change

This captures the various ways that node-gyp can find its Python executable within the entrypoint scripts and forwards them to the Python interceptor script as `ATOM_APM_ORIGINAL_PYTHON`. python-interceptor then execs `ATOM_APM_ORIGINAL_PYTHON` instead of a hardcoded "python".

Furthermore, this deals with some odd edge cases related to `npm` being _case-insensitive_ when it reads `npm_config_*` variables from its environment. If a user had something like `NPM_CONFIG_PYTHON` or `npm_CONFIG_node_GYP` set in their environment the results would technically be undefined. 🌈 

### Alternate Designs

I considered using `npm config set` and `npm config get` to use npm itself for precedence rules, but `npm config get` reported "undefined" for unset variables instead of a blank string (and returns with a 0 status), which makes it awkward to use in scripts.

### Benefits

#673 introduced a regression: apm no longer respects a user's external `PYTHON` or `npm_config_python` settings. This causes problems on systems where the default Python executable is not compatible with node-gyp. With this change, those settings are respected in a way which is consistent with the behavior of a plain`npm install`.

### Possible Drawbacks

This uses bash syntax pattern substitution constructs like `${..##}` and `${..%%}` which may not be portable to all shell environments. The shebang in our `apm` and `npm` scripts was already `#!/bin/bash`, though, so we should be no less portable than we were before.

### Applicable Issues

Fixes #698.

Issue #698, opened 2017-03-14
Python2 regression with 1.15.3


Hufschmidt:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

## Description
It seems like b364c3cfc1e28fce66418266bc719d08e614baf9 introduced a regression for Arch-Linux users, as node-gyp tries to use python3 instead of python2 when beeing executed via apm.

This is the same issue as with #95, but due to the aforementioned commit the workarounds `export PYTHON=python2` or `npm config set python /usr/bin/python2 -g` do no work anymore.

### Steps to Reproduce

**Method 1**
1. Boot into an arch-linux system (*eg. [Manjaro LiveCD](https://manjaro.org/get-manjaro/)*)
2. Install atom-editor from AUR (*Version 1.15.0 at the time of this writing*)
3. Install git-blame via apm (*which uses node-gyp*)

**Method 2**
1. Temporary symlink `/usr/bin/python` to `/usr/bin/python3` instead of `/usr/bin/python2`.
2. Try (re-)installing git-blame via apm (*which uses node-gyp*)

**Expected behavior:** git-blame installs

**Actual behavior:** git blame fails with

```
$ apm install git-blame
Installing git-blame to /home/hufschmt/.atom/packages ✗

> runas@3.1.1 install /tmp/apm-install-dir-117214-7034-190qczt/node_modules/git-blame/node_modules/runas                                                                                                               
> node-gyp rebuild                                                                                                                                                                                                     

/tmp/apm-install-dir-117214-7034-190qczt                                                                                                                                                                               
└── (empty)                                                                                                                                                                                                            


gyp ERR! configure error                                                                                                                                                                                               
gyp ERR! stack Error: Python executable "/usr/share/atom/resources/app/apm/bin/python-interceptor.sh" is v3.6.0, which is not supported by gyp.                                                                        
[...]

PR #875, opened 2019-11-17
Upgrade npm

smashwilson:
### Requirements

### Description of the Change

This upgrades our bundled npm from 6.2.0 to the latest release, npm 6.13.0.

### Alternate Designs

Because my primary motivations for doing this are to take advantage of node-gyp improvements, I thought about adding a direct dependency on node-gyp, which would let us control our node-gyp version directly moving forward. However, relying on npm's node-gyp dependency means that we benefit from npm's diligence in vetting compatibility and so on, and makes us less likely to neglect to upgrade node-gyp as well as npm in the future.

### Benefits

The newer version of node-gyp introduced by this bump - [v5.0.5](https://github.com/nodejs/node-gyp/blob/master/CHANGELOG.md#v505-2019-10-04) - features Python 3 compatibility and support for VS2017 and VS2019. Significantly, this will unblock us from being able to test Atom packages on GitHub Actions, whose Windows runners include (at the oldest) [VS2017](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners#windows-server-2016). This should unblock me on https://github.com/atom/github/pull/2298, although I won't know for sure 

The newer node-gyp also includes an upstream fix for the hack we used to #499 forever ago, introduced in #673 and #703. This should address Python-related issues like #818 and #698.

Beyond that, this will also allow us to take advantage of other npm improvements as found in [their CHANGELOG](https://github.com/npm/cli/blob/latest/CHANGELOG.md#6130-2019-11-05).

### Possible Drawbacks

This is a minor-version dependency bump and the npm client is used by far more people than those who use apm, so it should be quite safe. There's always a risk that we'll break some package or other though.

### Verification Process

We have good unit test coverage for the problem described in #499.

### Applicable Issues

#499, #818, #698, likely others.


DeeDeeG:
Hi, I was looking into migrating `atom` to Python 3 and found this PR.

I see that this PR currently breaks the tests for this repo, but the built `apm` seems to work properly.

Steps to reproduce: 
- Check out this branch
- `npm install` to install dependencies and build `apm`
- `./bin/apm install tree-view`

Explanation/results of steps to reproduce: 

You can run `./bin/apm install [some-package]` e.g. `./bin/apm install tree-view` and it works on this branch. (i.e. it successfully installs `tree-view` to `~/.atom/packages/tree-view`). Works with all typical Python configurations, i.e. with `python` symlinked to `python2`, `python` symlinked to `python3`, or no versionless `python` whatsoever (assuming there is still at least one of `python2` or `python3` on the `PATH`). Tested on Ubuntu 20.04 Beta.

Note:

I can confirm this PR is an improvement over stable/release `apm`, or `master` of this repo, as they both do not support `python3`.

(In stable `apm` 2.4.3 as bundled with `atom` 1.45.0, trying to do `apm install tree-view` only works only with `python` symlinked to `python2`. It errors out with `python` symlinked to `python3`, or with no versionless `python` whatsoever.)

See also:

My minimal branch with similar results to this PR (working `apm` binary, but broken tests): https://github.com/atom/apm/compare/master...DeeDeeG:update-npm

smashwilson:
Yeah, I think the CI on this repo is just haunted at this point 😅 Although the failures on this branch are different from the ones you can get on master, I noticed.

DeeDeeG:
I made a branch with only 6 test failures: https://travis-ci.com/github/DeeDeeG/apm/builds/161159412

The main difference is dropping the outdated `jasmine-focused` wrapper that depended an old fork of `jasmine-node`, and instead using up-to-date `jasmine-node` directly.

That passes ci testing when implemented against `master` (See: https://travis-ci.com/github/DeeDeeG/apm/builds/161139015), so I believe that indicates that the basic premise of performing that update is sound.

As such, I hope these last six test failures against this PR's branch are mostly legitimate indications of a small amount of work left to be done to tidy up for this PR, whether it be just fixing the tests, or tracking down genuine bugs.

DeeDeeG:
Short summary of this comment: I know what's breaking the tests on Python 2, and have patches to fix all but four of the failures (in two very similar/almost identical tests) under Python 2. Many more tests fail on Python 3, and I'm stumped for the moment on how to fix those.

---

I found out what the issues are that are breaking the tests. (At least under Python 2). Variations on "Node isn't sure what package to use, or where to find it."

- npm "deduped" away a version of `node-gyp`, such that the version under the npm directory (`node_modeules/npm/node_modules/node-gyp`) isn't populated, but the general/shared version (`node_modules/node-gyp`) is still populated. A test hard-coded the former path, which is empty now, so `node-gyp` isn't found, so it doesn't run, and the test fails.
  - Solution: Change hard-coded `node-gyp` path in the tests from the former path to the latter one. See [this commit](https://github.com/DeeDeeG/apm/commit/000d5f85e3ea2b58cf7aebc7fe5fcb63ae496d85) that fixes the issue.
  - Potentially better solution: Find a way to dynamically use `node-gyp` wherever it's actually located, don't use a hard-coded path.
- The tests seem confused about which version of `tar` to use. Part of the tests end up calling the 2.x version of `tar` with syntax that would have worked for version 4.x of `tar`, but it's the wrong major version. There was a [breaking change in command-line syntax](https://github.com/npm/node-tar/blob/master/CHANGELOG.md#30) starting with version 3.x of `tar`. Can't figure out what went wrong exactly to cause this???
  - Solution: Get npm to record better info in `package-lock.json`
    - I can't figure out what steps to do to make this happen, but I was able to do it several times and [commit](https://github.com/DeeDeeG/apm/commit/fa3f7c307c9b946ec09a36d8f8369d8f809b9f8b) it, so... Problem solved, for now?
  - Even better solution: Use the same major version of `tar` as `npm` uses, so node doesn't have to guess which version of `tar` to use among two mutually-incompatible major versions. (npm is using `tar` 4.x at the moment. I have [a commit](https://github.com/DeeDeeG/apm/commit/484d1e158b077d68cc05b7d23cd90190370d2fae) to switch to that if folks here would like to do that.)

There's another issue as well (I don't have a patch for this at the moment; I'm not very fluent with JS so it's a lot to pick up at once here):
- Newer `node-gyp` verifies that your Python binary is an actual Python binary. So the [dummy](https://github.com/atom/apm/blob/707cd2a8ab434358c1940640c3f28bd44d71bce2/spec/fixtures/fake-python-1.sh) [files](https://github.com/atom/apm/blob/707cd2a8ab434358c1940640c3f28bd44d71bce2/spec/fixtures/fake-python-2.sh) we use are ignored, and the tests that check if you can configure a custom Python path fail. The feature works in `node-gyp`, but we would need to supply a real Python binary as the dummy file to have these tests work. Newer `node-gyp` also formats its errors differently now, so the tests need to be re-written to account for that.
  - Solution: Re-write [these tests](https://github.com/atom/apm/blob/707cd2a8ab434358c1940640c3f28bd44d71bce2/spec/install-spec.coffee#L570-L610) in a way that works with newer `node-gyp`.

These fixes are enough to make all the tests pass using Python 2.7, but several more tests fail when using Python 3 (I'm using Python 3.8.2 for my version of python3). So there would be additional work to track those down. Most/all of these seem somehow obscurely related to `node-gyp`, but I can't say for sure. I am tempted to try updating all the old pinned dependencies in this repo and see if that fixes the test failures under Python 3.

---

There's another issue that tends to get fixed automatically by running enough `npm install` commands:

- A misformed git dependency URL was committed somehow in `package-lock.json`.

```console
Could not install from "node_modules/jasmine-focused/jasmine-node@git+https:/github.com/kevinsawicki/jasmine-node.git#81af4f953a2b7dfb5bde8331c05362a4b464c5ef" as it does not contain a package.json file.
```

  - Solution: Find and replace `jasmine-node@git+https://github.com` with `git+https://github.com` in `package-lock.json`

smashwilson:
Ooh thanks for taking the time to dig in further, @DeeDeeG  😍 

Can you create a new pull request with the commits from this ref and yours? You're a lot closer than I was, and I likely won't have time to do a bunch of cherry-picking to take this one over the edge before you could.

DeeDeeG:
I posted #887 with the minimal-diff/less-adventurous versions of the fixes.

Issue #328, opened 2015-03-26
apm does not (appear) to have the same node version as atom


Arcath:
When installing [node-sass](https://github.com/sass/node-sass) it compiles a binding for your os/node version which it then calls later when you require it.

When install via `apm install` it creates (on my machine) `win32-ia32-node-0.10` which is rightly not my system node version. If i then try to require `node-sass` in my package I get the error that no binding is available and I should re-install node-sass.

Looking at the `process.sassBinaryName` in Atom it is looking for `win32-ia32-atom-1.0`

Should apm not be using an identical environment to Atom so that any module that compiles anything will work when required in Atom?

kevinsawicki:
> Should apm not be using an identical environment to Atom so that any module that compiles anything will work when required in Atom?

apm compiles all native modules against the v8 headers that ship with Atom.

What error are you getting when `node-sass` is required?

Arcath:
The error is

`libsass` bindings not found. Try reinstalling `node-sass`

Arcath:
Doing some more testing with the node-sass 3.0.0 beta which uses `{platform}-{arch}-{modules version}` to name its libsass binding. Installed by apm on my machine that results in `win32-ia32-11` but run the same function in atom and I get `win32-ia32-43`.

running `process.versions` in atom: 
```
modules: '43'
node: '1.5.1'
...
v8: '4.1.0.21'
```

run the command `C:\path\to\atom\app-0.192.0\resources\app\apm\bin\node -e 'console.log(process.versions);'`
```
http_parser: '1.0',
node: '0.10.35',
v8: '3.14.5.9',
ares: '1.9.0-DEV',
uv: '0.10.30',
zlib: '1.2.8',
modules: '11',
openssl: '1.0.1j'
```
which returns a vastly different answer and using that information would give the binding I get during install.

kentonv:
> apm compiles all native modules against the v8 headers that ship with Atom.

Sure, but many npm modules include build scripts that do things like try to stamp the library with the node or v8 version to detect ABI version skew. Those scripts make the reasonable assumption that they are building for the version of node which is executing the script. But under apm, these scripts are run using node 0.10, so they get confused and use that version for the stamp, even if they are actually building against newer headers.

    $ /usr/share/atom/resources/app/apm/bin/node --version
    v0.10.40

Why is Atom shipping this old node binary when Atom proper is running on something much newer?

50Wliu:
[...]

PR #896, opened 2020-08-19
node 12.4.0 - Electron 6

aminya:
by @DeeDeeG:

### Description of the Change

Update bundled Node to v12.4.0, which matches [the version in Electron 6](https://www.electronjs.org/releases/stable?version=6&page=6).

### Alternate Designs

None. We intend to match the Node version in `apm` to that which the version of Electron we use in Atom is based on.

Our Atom is using Electron 6. So we should bundle Node v12.4.0 in `apm`.

### Benefits

For packages that try to detect the version of Node they are building against during install, this closely approximates the version of Node used in Atom's version of Electron. So if there are any potential bugs in various packages due to non-matching versions of Node/Electron, this should take care of 99% of them, maybe even all of them.

### Possible Drawbacks

None expected. This is just a minor version update of Node.

### Verification Process

CI and manual testing, to be done soon.

### Applicable Issues

For a discussion of the type of issue this is designed to address, see: https://github.com/atom/apm/issues/328


darangi:
Thanks @aminya

PR #897, opened 2020-08-19
.gitattributes: Set platform-specific line-endings

### LINKED ISSUE #7: Extract atom package classes from atom repo
apm should be the home of things like `AtomPackage`, `Package`, and much of the package related code in the `atom` global.  These currently still sit in the `atom/atom` repository.  Also there is a `PackageManager` class in atom/settings-view that should probably be pulled in as well into apm.

Pull these out once atom/atom#775, atom/atom#778, and atom/atom#787 land.

aminya:
By @DeeDeeG 
### Description of the Change

- Use the [`.gitattributes`](https://www.git-scm.com/docs/gitattributes#_eol) file to ensure:
  -  `.cmd` files are checked into/checked out of the git object store with Windows-friendly `CRLF` line endings.
  - bash scripts use Unix-friendly `LF` line endings.

See: https://en.wikipedia.org/wiki/Newline#Issues_with_different_newline_formats

### Alternate Designs

None.

### Benefits

With this change, users should get working scripts as part of their `apm` install, regardless of what platform the module was published from.

(Wrong line endings can cause scripts not to run. It appears that without this PR, files may be checked out locally with wrong line endings, and subsequently published to the npm package registry with wrong line endings.)

(This PR expected to prevent platform-specific files from being published to the npm package registry with the wrong line endings.)

### Possible Drawbacks

None.

### Verification Process

Changing the line endings of the bash scripts to `LF` allowed them to run on my Linux machine.

Presumably setting the `.cmd` files to have Windows line-endings has a similar benefit.

### Applicable Issues

https://github.com/atom-ide-community/apm/issues/7 has a description of the problem this fixes. (This PR closes https://github.com/atom-ide-community/apm/issues/7).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,7 @@
+# Shell scripts (Unix-style line endings)
 *.sh    text eol=lf
+bin/apm text eol=lf
+bin/npm text eol=lf
+
+# Batch files (Windows-style line endings)
+*.cmd   text eol=crlf


DeeDeeG:
@aminya please update the Applicable issues to link to our fork's issue `#7`. Or I suppose we can open an equivalent issue here, but I think linking to our issue is good enough.

aminya:
I linked the correct issue. Recreation seems redundant.

DeeDeeG:
Better summary of why this PR is good: _This prevents Windows users from publishing the package's shell scripts (such as `bin/apm` and `bin/npm`) with Windows line-endings; Windows line-endings break the shell scripts._

(I was the one that wrote the complicated PR text. Sorry I tend to explain things too technically when the info can be much simpler).

PR #898, opened 2020-08-23
use node-gyp correctly in the tests

aminya:
### Description of the change

This uses npm's built-in node-gyp which is used in the tests. The tests assume that node-gyp should exist in node_modules, but this is an assumption about the test environment which may not be true. npm does not hoist `node-gyp` which is a dependency of the `npm` to the top-level `node-modules` in the newer versions. It adds `node-gyp` to devDependencies just for `install-spec` test.

https://github.com/atom/apm/blob/475d35f92b0e257bb0c3fdbdb8e302ee8fb0d2bb/spec/install-spec.coffee#L536-L541

### Verification 
The tests now pass correctly. This is tested on Atom too: https://github.com/atom-ide-community/atom/pull/103

### Drawbacks
none

### Release Notes
N/A

Closes #10

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,6 @@
     "grunt-contrib-coffee": "^2.0.0",
     "grunt-shell": "^1.3.0",
     "jasmine-focused": ">=1.0.7 <2.0",
-    "node-gyp": "^7.1.0"
+    "node-gyp": "^5.1.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "grunt-coffeelint": "0.0.16",
     "grunt-contrib-coffee": "^2.0.0",
     "grunt-shell": "^1.3.0",
-    "jasmine-focused": ">=1.0.7 <2.0"
+    "jasmine-focused": ">=1.0.7 <2.0",
+    "node-gyp": "^7.1.0"
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1575,8 +1575,7 @@
     "env-paths": {
       "version": "2.2.0",
       "resolved": "https://registry.npmjs.org/env-paths/-/env-paths-2.2.0.tgz",
-      "integrity": "sha512-6u0VYSCo/OW6IoD5WCLLy9JUGARbamfSavcNXry/eu8aHVFei6CD3Sw+VGX5alea1i9pgPHW0mbu6Xj0uBh7gA==",
-      "dev": true
+      "integrity": "sha512-6u0VYSCo/OW6IoD5WCLLy9JUGARbamfSavcNXry/eu8aHVFei6CD3Sw+VGX5alea1i9pgPHW0mbu6Xj0uBh7gA=="
     },
     "err-code": {
       "version": "1.1.2",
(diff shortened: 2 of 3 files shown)


DeeDeeG:
For what it's worth, the `require.resolve('node-gyp')` workaround is all that should be needed here, IMO. That should work without including `node-gyp` in `devDependencies`. (I originally coded this type of fix myself some time ago, and it worked without adding `node-gyp` to `devDependencies`.)

Pinning an exact version of `node-gyp` in `devDependencies` should make this test pass forever. That's true. But it would be pretty stable even without `node-gyp` in `devDependencies` too. It would work without `node-gyp` in our `devDependencies` so long as the entrypoint of the `node-gyp` module (`node-gyp/lib/node-gyp.js`) doesn't move to a different depth of folders within the module. That is to say, the `node-gyp` maintainers would have to seriously refactor their module for this test to break, even without pinning `node-gyp` in `devDependencies`.

`node-gyp`'s install location within `apm`'s `node_modules` is somewhat variable, because multiple dependencies we have require `node-gyp` as a sub-dependency. During `npm install`, sometimes the required version ranges all match, and they can all be deduped. Other times ther version ranges are different, and our `npm` dependency needs its own copy. Stuff like that. It doesn't necessarily move strictly because of something the `npm` authors did (if anything) to change `npm`.

aminya:
> For what it's worth, the `require.resolve('node-gyp')` workaround is all that should be needed here, IMO. 

That's not correct. Using a newer npm (#900) changes the location of node-gyp which results in the test failure.

package.json
```diff
@@ -56,6 +56,7 @@
     "grunt-coffeelint": "0.0.16",
     "grunt-contrib-coffee": "^2.0.0",
     "grunt-shell": "^1.3.0",
-    "jasmine-focused": ">=1.0.7 <2.0"
+    "jasmine-focused": ">=1.0.7 <2.0",
+    "node-gyp": "^7.1.0"
```
DeeDeeG, line 60: - IMO `node-gyp` in `devDependencies` should be set to an exact version, (no caret `^` semver range) so it's stable forever.

- Or it can be removed from `devDependencies`, and just allow `require.resolve('node-gyp')` to find the indirect dependency from `npm` --> `node-gyp`.

DeeDeeG:
> That's not correct.

I don't mean to argue for no reason. But I did try this. It passed tests. I'm not just pulling it from thin air. If you look in the commits for this very PR, the commit just before adding `node-gyp` to `devDependencies` also passes CI. https://github.com/atom/apm/pull/898/commits/622631421b2f0d1726d88125686798886c99a83e

Meaning no disrespect. Just giving a technical review.

aminya:
> also passes CI. [6226314](https://github.com/atom/apm/commit/622631421b2f0d1726d88125686798886c99a83e)

This is with the old npm.

I have already tested this: https://github.com/atom-ide-community/apm/pull/10#issuecomment-678738490

> We need to install node-gyp for testing apm on the newer npm versions. There is no way around this, and we should not rely on the hoisting algorithm.

DeeDeeG:
I see you are right. Sorry for the misunderstanding.

DeeDeeG:
Another small thing: There is missing `node-gyp@5` on this branch. I'm not sure why, but there is no `node-gyp@5` in the `package-lock.json` even though `npm` is supposed to require that.

It got deduped a little too aggressively, I guess. (That would mean we will test on `node-gyp` 7, but Atom would actually run `node-gyp` 5. We should test the same thing we use in production.)

The fix is to run `npm install --save-dev node-gyp` again. That will re-add `node-gyp@5` that `npm` needs.

Do you mind if I add it to the branch for this PR? Or do you want to do it?

aminya:
I don't understand what you are trying to achieve. I just added a devDependency which is only used in the tests. This does not affect Atom or anything. This machine-generated package-lock.json does not change by running `npm install`.

DeeDeeG:
It's just a weird bug with how the `package-lock.json` got saved. (I swear one out of every ten pull requests related to Atom or apm has had something like that come up as an issue. :frowning_face:.) `package-lock.json` on this branch was saved in such a way that all CI runs, and any developers hacking on `apm`, are ONLY going to get one copy of `node-gyp`, and it will be version 7.

Here's the output from `npm ls node-gyp` on this branch without the fix:

```console
$ npm ls node-gyp
atom-package-manager@2.5.2 /home/mate/apm
├── node-gyp@7.1.0 
└─┬ npm@6.14.4
  ├── UNMET DEPENDENCY node-gyp@^5.1.0
  └─┬ npm-lifecycle@3.1.4
    └── UNMET DEPENDENCY node-gyp@^5.0.2

npm ERR! missing: node-gyp@^5.1.0, required by npm@6.14.4
npm ERR! missing: node-gyp@^5.0.2, required by npm-lifecycle@3.1.4
```

With the fix:

```console
$ npm ls node-gyp
atom-package-manager@2.5.2 /home/mate/apm
├── node-gyp@7.1.0 
└─┬ npm@6.14.4
  ├── node-gyp@5.1.1 
  └─┬ npm-lifecycle@3.1.4
    └── node-gyp@5.1.1 
```

With `node-gyp@5` missing, all tests will run with `node-gyp@7`. That is slightly different from what end-users will actually get bundled with Atom.

(node-gyp 5 and 7 are mostly very similar... but they are a bit different. In the rare situation where they behave differently, we don't want CI to pass when it should have failed. And we don't want it to fail when it should have passed. So we need the copy of `node-gyp@5` that's nested under `npm`.)

This is about testing with the same code that we will ship.

aminya:
The point of this PR is to not rely on npm's hoisting algorithm and package-lock.json which are susceptible to change. Updating package-lock.json with a different command is not sustainable and is a very fragile approach.

 If a certain version of `node-gyp` needs to be a devDependency it should be declared in package.json. I added version 5 just in case Atom's build scripts rely on this fragile behavior.

DeeDeeG:
That works. Thanks for addressing it. This look good to me now.

DeeDeeG (approved):

darangi:
Thanks 🙇🏾‍♂️  @aminya

aminya:
You're welcome!

PR #916, opened 2021-03-06
feat: update the bundled Node version to match Atom-Electron 9

aminya:
### Description of the Change

feat: update the bundled Node version to match Atom-Electron 9

https://www.electronjs.org/blog/electron-9-0

### Benefits
Matching the same Node version means more stability across the whole Atom ecosystem.

### Verification Process
The tests/CI pass


BUNDLED_NODE_VERSION
```diff
@@ -1 +1 @@
-v12.4.0
+v12.14.0
```
aminya, line 1:
```suggestion
v12.14.1
```

aminya:
@sadick254 @darangi This again has caused many issues for us. Linux users are reporting failures for installing Hydrogen while everything is already provided. Apm detects the ABI version as 72 while Atom uses Electron 9 which is ABI 80! Could you please merge this as soon as possible?

https://github.com/nteract/hydrogen/issues/2091#issuecomment-821977900

Issue #908, opened 2020-10-28
Cannot publish a new package when the default branch is called "main" instead of "master"


aminya:
<details>
<summary>Prerequisites</summary>

### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: <https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode>
    * Followed all applicable steps in the debugging guide: <https://flight-manual.atom.io/hacking-atom/sections/debugging/>
    * Checked the FAQs on the message board for common solutions: <https://discuss.atom.io/c/faq>
    * Checked that your issue isn't already filed: <https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom>
    * Checked that there is not already an Atom package that provides the described functionality: <https://atom.io/packages>

</details>

### Description

When the default branch of a repository is called main (this is currently the default behavior of GitHub), you cannot publish the apm package.

This is the kind of error that will be thrown:
```
❯ apm publish minor
Registering atom-ide-console failed
Registering package in atom-ide-community/atom-ide-console repository failed: That repo does not exist, isn't an atom package, or you do not have access
```

Manually tagging and using `apm publish --tag "v0.1.0" does not help either. 

The only solution I have found is to make a branch from the main and call it "master". Make it the default branch in the GitHub settings, and then publish the package.

### Steps to Reproduce

1. Create a new GitHub repository and make the default branch name "main"

2. Create an atom package in the same repository
```
apm init --package .
```

3. Try to publish
```
apm publish patch
```

**Expected behavior:**

Publish 

**Actual behavior:**

See the above description.

**Reproduces how often:**
Always

### Versions
```
❯ apm --version
apm  2.5.2
npm  6.14.8
node 12.4.0 x64
atom 1.52.0
python 2.7.9
git 2.29.1.windows.1
visual studio
```

1000p:
https://discuss.atom.io/t/apm-publish-error/79274/2 I can confirm

PR #917, opened 2021-03-06
Fix/first publish master needed

Nicolas-Reyland:
### Requirements

### Description of the Change

When publishing an atom package through the apm, if the repository is set on the default "main" branch (new git standard), you would get an error. This is because the publish script is looking for a "master" branch (old standard):
 - (line 252) checks if the repo is set to an "alternative" (not the main) branch, if one is found, accesses it
 - (line 254) it then tries to access the "master" (hard-coded 'branch.master.remote')

The new git convention is that the main branch should be named "main". Instead of checking for the master branch, I first check if there is a "main" branch, then I check if there is a "master" branch.

If I got something wrong, please tell me: this is my first ever pull request. My request might be bad, and I am aware that you normally don't tell people when their requests were rejected (it is written in this text file, in the comments), but please tell me why this might be the case, so I know how to do them in the future

### Alternate Designs

You could also only check for the main branch, not checking for the master branch afterwards.

### Benefits

You will not get the error saying the git repo is not valid (not an atom package, not sufficient permission or it doesn't exist)

### Possible Drawbacks

I cannot think of a possible drawback.

### DIFF
--- a/src/publish.coffee
+++ b/src/publish.coffee
@@ -252,6 +252,7 @@ class Publish extends Command
     if currentBranch = repo.getShortHead()
       remoteName = repo.getConfigValue("branch.#{currentBranch}.remote")
     remoteName ?= repo.getConfigValue('branch.main.remote')
+    remoteName ?= repo.getConfigValue('branch.master.remote')
 
     upstreamUrl = repo.getConfigValue("remote.#{remoteName}.url") if remoteName
     upstreamUrl ?= repo.getConfigValue('remote.origin.url')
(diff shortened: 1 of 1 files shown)


Nicolas-Reyland:
This is my first pull request. If this pull request is rejected, please tell me why !
I forgot to add: I rebuilt everything and it succeeded.

aminya:
Does this fix #908

Nicolas-Reyland:
> Does this fix #908?

It is supposed to fix #908 , yes. To be completely honest, I recompiled (using your instructions in the README), and everything recompiled smoothly, but I did not know how to use the version of apm I compiled (the bin/ directory content did not change after compiling?).
How can I test it ?

Nicolas-Reyland:
Doesn't fix. sry

idleberg:
I think `main` should be the new default, but how about a command-line flag that allows specifying any other branch

